PR #24, opened 2015-03-17
Issue7 : Simplify Attribute API

rnc:


rnc:
For issue #24

jdcasey:
I guess I'd been thinking of more a composition approach, where Attributes is a simple class that stored a map of attributes and provided constant keys to retrieve them. It might provide the most basic convenience methods like getTimeoutSeconds() (but thinking about that more, I'm not sure it's a good idea).

Thinking about this more, I'm starting to wonder why we don't just make the attribute map an implementation pattern, along with get/setAttribute(..), that gets implemented in the appropriate places. I'm not sure there's much to be gained by formalizing it in its own class / base-class. I think I'd rather just simplify all this that way. There are only two places where it's really implemented.

jdcasey:
I've committed an alternative implementation that just puts the attribute map in the Location impl directly and removes the confusing getTimeoutSeconds() methods...then consolidated the constants used to access the attributes into Location.

If we find later that we need to factor out the attribute stuff, we can, but it seems like overkill at the moment.

PR #33, opened 2015-05-18
adding pompeek class

jdcasey:


rnc:
Is this the same PomPeek as from PME ? Should we migrate this to be the canonical location (including the unit tests) ?

jdcasey:
yeah, unit tests are a great point. It's the same (with a couple minor additions)

Issue #58, opened 2015-07-29
Galley should provide more information about transfer failure modes


jdcasey:
especially timeouts. It should be possible to query the TransferException to find out why it failed.

PR #59, opened 2015-07-30
Adding timeout/bad-gateway exception types, and adjusting thrown exce…

jdcasey:
…ptions accordingly. Configuring timeout on http requests, and doing a more careful job analyzing the type of response to determine the error to throw.


jdcasey:
should solve #58

PR #68, opened 2015-09-18
Adding module to help with embedding in a CDI application. This inclu…

jdcasey:
…des galley-maven, http transport, and file cache provider.


jdcasey:
retest this please

PR #76, opened 2015-11-10
adding inherited/mixin status to POM element views, switch to jhttpc to back http transport

jdcasey:


jdcasey:
retest this please

PR #91, opened 2016-02-18
Fix SpecialPathConstants to avoid multi-level metadata again

pkocandr:
The checksum regex patterns were not specific enough. The checksums
metadata also matched the checksum pattern, so metadata got all the
properties of checksums (listable, publishable, retrievable, storable)
when they were checked against the checksum pattern first.

It led to creation of additional metadata levels with each listing and
in the end to "filename too long" exceptions.

Example of such filename found in Indy log:
/org/codehaus/woodstox/stax2-api/3.1.4/stax2-api-3.1.4.pom.sha1.http-metadata.json.asc.http-metadata.json.asc.http-metadata.json.md5.http-metadata.json.sha1.http-metadata.json.sha1.http-metadata.json.asc.http-metadata.json.md5.http-metadata.json.sha1.http-metadata.json.sha1.http-metadata.json.md5


pkocandr:
This is a foolow-up of #88 which did not fix all the problems.

Issue #94, opened 2016-03-11
Provide access to Properties in MavenPomView


rnc:
Also useful in the context of https://github.com/release-engineering/pom-manipulation-ext/issues/319

jdcasey:
I think this is implemented, though we may need to revise the implementation later. Please reopen if that's not accurate.

PR #95, opened 2016-03-14
Issue #94 Provide access to Properties

rnc:


maven/src/main/java/org/commonjava/maven/galley/maven/model/view/PropertiesView.java
```diff
@@ -0,0 +1,110 @@
...
+
+        for ( Element e : elements )
+        {
+            Node value = e.getFirstChild();
+            result.setProperty( e.getNodeName(), (value == null ? "" : e.getFirstChild().getNodeValue() ) );
+        }
+
+        for ( String key : result.stringPropertyNames() )
+        {
+            if ( result.getProperty( key ).startsWith( "${" ) )
```
jdcasey: Try:
```
    String value = getPomView().resolveExpressions(result.getProperty(key)); //resolves maven expressions
```

PR #99, opened 2016-04-22
Remote repo contents filtering

pkocandr:


jdcasey:
@pkocandr +1, this looks good to me. Are you ready to merge it, or are you still tweaking it?

pkocandr:
I'm not tweaking this anymore. I think it would be very useful to have some general repo-filtering solution, but we probably don't have enough resources to do that now.

jdcasey:
Okay. I'll merge this now, and if we have something that supercedes it in Indy later, I guess we can talk about whether or not to deprecate this. Thanks!

PR #101, opened 2016-04-27
add a metadata timeout propeties in Location Interface and SpecialPat…

ligangty:
Used for indy's metadata timeout function


jdcasey:
Add to whitelist

PR #106, opened 2016-05-06
Disable decoration of metadata checksums

pkocandr:
When a metadata checksum, e.g. maven-metadata.xml.md5, was decorable,
it created another level of checksums, e.g. maven-metadata.xml.md5.sha1
in group's storage and these extra checksums were also present in a
directory listing.

### DIFF
--- a/api/src/main/java/org/commonjava/maven/galley/io/SpecialPathConstants.java
+++ b/api/src/main/java/org/commonjava/maven/galley/io/SpecialPathConstants.java
@@ -17,13 +17,9 @@ package org.commonjava.maven.galley.io;
 
 import org.commonjava.maven.galley.model.FilePatternMatcher;
 import org.commonjava.maven.galley.model.SpecialPathInfo;
-import org.commonjava.maven.galley.model.SpecialPathMatcher;
-
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 
 /**
  * Created by jdcasey on 1/27/16.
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/commonjava/maven/galley/io/SpecialPathConstants.java
```diff
@@ -56,20 +52,36 @@
...
                             .build();
 
         sp.add( pi );
 
-        pi = SpecialPathInfo.from( new FilePatternMatcher( "archetype-catalog\\.xml(\\.md5|\\.sha[\\d]+)?$" ) )
+        pi = SpecialPathInfo.from( new FilePatternMatcher( "maven-metadata\\.xml(\\.md5|\\.sha[\\d]+)$" ) )
+                .setDecoratable( false )
+                .setMergable( true )
+                .setMetadata( true )
+                .build();
```
jdcasey: Hm, these aren't mergable...

pkocandr: How so? You need merged metadata file to get the checksum, not first checksum you find in a constituent.

jdcasey: I just meant that we wouldn't actually merge these files...we'd generate a new one for the merged metadata file they are derived from. I'm not sure it matters in practice...but it's not straightforward to create one of these for a group if it's the first thing you request. You have to generate the metadata file it's derived from, then generate it.

pkocandr: You are right, but in this case the meaning of the flag is "in case of group don't pick the first matching file from a constituent but rather use a generator to produce a result from all constituent, not necessarily from files with the same name". Maybe the name of the flag should be different, but can't find anything short and fitting to match the above meaning :-) In any case I believe the value should be true to work properly.

PR #117, opened 2016-07-21
NOS-13, De-duplicate overlapping elements in Galley Maven POM view results

yma88:
De-duplicate overlapping elements in Galley Maven POM view results


jdcasey:
add to whitelist

PR #119, opened 2016-07-26
Implement NFS storage CacheProvider

ligangty:
Hi @jdcasey . I've fixed most of your comments for last commit, but still some stuff here:

For openOutputStream, you mentioned that we need some idle timeout way in the stream wrapper to handle lock releasing for stream no operation in seconds, I'm still confused how to do this? As I understand, when the txMgr.commit/rollback, the lock will be released. I've added these code snippet in the wrapper's close() in this commit. Need some clues to do this idle if this code snippet can not work.

And for the waitForXLock, I'm using the lock owner in ISPN cache to wait, which I think it can be better than the thread.sleep.

### DIFF
--- a/caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
+++ b/caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
@@ -1,0 +1,623 @@
+package org.commonjava.maven.galley.cache.infinispan;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.StringUtils;
+import org.commonjava.cdi.util.weft.ExecutorConfig;
+import org.commonjava.cdi.util.weft.WeftManaged;
+import org.commonjava.maven.galley.cache.partyline.PartyLineCacheProvider;
+import org.commonjava.maven.galley.model.ConcreteResource;
+import org.commonjava.maven.galley.model.Location;
+import org.commonjava.maven.galley.model.Transfer;
+import org.commonjava.maven.galley.spi.cache.CacheProvider;
+import org.commonjava.maven.galley.spi.event.FileEventManager;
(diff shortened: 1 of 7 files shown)


jdcasey:
add to whitelist

caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
```diff
@@ -0,0 +1,623 @@
...
+    @Override
+    public void stopReporting()
+    {
+        plCacheProvider.stopReporting();
+    }
+
+    @Override
+    public void cleanupCurrentThread()
+    {
+        plCacheProvider.cleanupCurrentThread();
```
jdcasey: WRT expiring streams, I think this is actually a good, reliable way to do it. We should keep a map `Map<threadId, Set<OutputStream>>` in this cache provider. When we open a new output stream wrapper to pass to the user, we should add it to the map using `Thread.currentThread().getId()` for the key. When this method (`cleanupCurrentThread()`) gets called, we should again grab the current thread's id, remove the corresponding set from the map, and iterate through it, closing each stream.

This means openOutputStream() and cleanupCurrentThread() will either have to be `synchronized` (or contain `synchronized` blocks), or else we might be able to get away with a ConcurrentHashMap...though if I remember correctly, that may be a pretty memory-hungry class.

The idea here is that after the user is done with a thread, some cleanup mechanism will run. In Indy, this happens in a servlet Filter, in a finally block (IIRC). All opportunity for user interaction is past, and the system cleans up any dangling resources related to that request.

For an example, see:

* https://github.com/Commonjava/partyline/blob/master/src/main/java/org/commonjava/util/partyline/JoinableFileManager.java#L67
* https://github.com/Commonjava/indy/blob/master/subsys/jaxrs/src/main/java/org/commonjava/indy/bind/jaxrs/ResourceManagementFilter.java#L71

ligangty:
Will fix that stream issue. Close this PR

PR #122, opened 2016-08-22
NOS-554 Refactor the current FastLocalCacheProvider with new design a…

ligangty:
…nd partyline


caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
```diff
@@ -63,6 +68,8 @@
 
     private String nfsBaseDir;
 
+    // use weak key map to avoid the memory occupy for long time of the transfer
+    //FIXME: should we mark this as static to make it used across all instances(for lock also)?
```
jdcasey: I think we can allow the application to manage the instances in the JVM...for instance, Indy uses initializes a single instance and uses a CDI provider to make it available. It's conceivable that we might want to use two of these in some weird system in the future. As long as they don't store references to the same file, it should work. OTOH, maybe there's something we should do to define a "root" within which this provider has responsibility for everything. I'm not sure how we could do that elegantly, TBH.

caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
```diff
@@ -93,26 +100,82 @@
     protected FastLocalCacheProvider()
     {
         nfsBaseDir = System.getProperty( NFS_BASE_DIR_KEY );
```
jdcasey: Once we have things closer to a final state, I'd like to replace this with a configuration object that we pass in. That gives us the option to enrich the config object without changing the API of the provider...and it allows us to avoid using static system properties.

jdcasey:
closing; #128 fixes merge conflicts and adds weak refs in the streams mapping

PR #123, opened 2016-08-22
NOS-447, Fix file locking in Galley's FileCacheProvider

yma88:
NOS-447, Fix file locking in Galley's FileCacheProvider


jdcasey:
@yma88 I think the only way to separate it is to use a different branch in your fork of the repository. However, this looks pretty good so far. I'm going to go over the 544-related stuff a bit more, but it builds and it looks pretty good on the surface. I'm content to clean it up with another PR if necessary, but I think we can proceed from here.

PR #129, opened 2016-08-31
NOS-556 Try to create some sorts of ftests based on byteman for FastL…

ligangty:
…ocalCacheProvider

Hi @jdcasey , I give a fix in this PR with the comment you mentioned yesterday about the openInputStream problem, and wrote a test ( testTryToWriteThenWaitReadingStreamOpenWithNFSExistedButLocalNot in FastLocalCacheProviderConcurrentTest). But seems  that when the readTask ran before the writeTask(which means the inputstream is got before the OuputStream.write) will cause unlimited lock. But when the write is before the read there is no this problem. I think this is because I did not got the right way to control the race condition in BMScript, so I disabled it. I'll continue to find the good way to control that in bmunit, and give this PR first to let you have a review. Waiting for your comments. Thanks.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -54,11 +54,6 @@
       <artifactId>weld-se-core</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.commonjava.cdi.util</groupId>
-      <artifactId>weft</artifactId>
-      <optional>true</optional>
-    </dependency>
-    <dependency>
       <groupId>org.commonjava.maven.galley</groupId>
       <artifactId>galley-test-harness-api</artifactId>
     </dependency>


caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
```diff
@@ -309,71 +309,111 @@ public boolean isFile( ConcreteResource resource )
...
+            @Override
+            public void run()
             {
-                final String pathKey = getKeyForResource( resource );
-                // will run the NFS->local copy in another thread, which can use PartyLine concurrent read/write function
-                // on the local cache to boost the i/o operation
-                executor.execute( new Runnable()
+                InputStream nfsIn = null;
+                OutputStream localOut = null;
+                synchronized ( copyLock )
```
jdcasey: I'm not sure if it matters, but I think you can wait to synchronize on the copyLock until you're ready to notify it. The other side, wait() will also need that synchronized block, but I'm pretty sure this whole section doesn't need to be synchronized that way (since the other side, of opening the input stream for the user, will wait until copyLock is notified).

This does bring up an interesting point, though. If another thread starts the openInputStream() process right after this one starts, but before the thread starts, it may try to spawn a second writer thread. I'm not sure whether this will result in an error, or corrupted data, or just one of the calls waiting longer for the first copy/read to complete.

I haven't had enough time to read up on ReentrantLock yet, but I'm wondering if maintaining a Map<String, ReentrantLock> field at the instance level might help here. We could use that in place of copyLock, maybe? I'm not 100% sure that makes sense. But it does feel like we're setting up a race condition here. I guess maybe a test would be a reasonable place to start pursuing that suspicion... :-\

caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
```diff
@@ -309,71 +309,111 @@ public boolean isFile( ConcreteResource resource )
...
+                        canStreamOpen.set( true );
+                        copyLock.notifyAll();
                     }
-                } );
+                }
+            }
+        };
+        // This lock is used to control the concurrent operations on the resource, like concurrent delete and read/write.
+        // Use "this" as lock is heavy, should think about use the transfer for the resource as the lock for each thread
+        synchronized ( getTransfer( resource ) )
```
jdcasey: Something tells me combining this synchronized block with the nested one on copyLock may be leading to the deadlock you mentioned in your comments at the top of the PR.

jdcasey:
@ligangty Thanks for filing the PR early. This gives us a chance to see what CI will do with it, so that helps (just in case the CI server is still having problems with bmunit, which it seems it's not). Also, it makes it a little easier to track changes I'm making with those you're making, I think.

PR #132, opened 2016-09-05
NOS-556 ftests and some fix for the FastLocalCacehProvider

ligangty:
Hi, @jdcasey 

In this new PR, I've narrowed the copyLock scope to notify action only. 
For your comments about the instance level map as the lock, I'm not sure, but I think it is a little heavy here as the lock. I had a think about this local variable copyLock and with the nest relation of the Transfer lock, and I can not think out the dead lock problem from code level. Of course the test for this is needed, and I'm still thinking how to simulate the dead lock case. If done, will push it later.
So just want to know your opinions about this.

### DIFF
--- a/caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
+++ b/caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
@@ -63,7 +63,8 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * This cache provider provides the ability to write the backup artifacts to an external storage(NFS) which can be mounted
@@ -87,7 +88,6 @@ public class FastLocalCacheProvider
     private String nfsBaseDir;
 
     // use weak key map to avoid the memory occupy for long time of the transfer
-    //FIXME: should we mark this as static to make it used across all instances(for lock also)?
     private final Map<ConcreteResource, Transfer> transferCache = new ConcurrentWeakKeyHashMap<>( 10000 );
 
     private final Map<String, Set<WeakReference<OutputStream>>> streamHolder = new HashMap<>( 50 );
@@ -113,7 +113,7 @@ public class FastLocalCacheProvider
     @ExecutorConfig( named = "fast-local-executor", threads = 5, priority = 2, daemon = true )
     @WeftManaged
     @Inject
-    private Executor executor;
+    private ExecutorService executor;
 
     private PathGenerator pathGenerator;
 
@@ -129,14 +129,14 @@ public class FastLocalCacheProvider
      *
      * @param plCacheProvider - PartyLineCacheProvider to handle the local cache files
      * @param nfsUsageCache - ISPN cache to hold the nfs artifacts owner
-     * @param fileEventManager
-     * @param transferDecorator
-     * @param executor
+     * @param fileEventManager -
+     * @param transferDecorator -
+     * @param executor -
      */
     public FastLocalCacheProvider( final PartyLineCacheProvider plCacheProvider,
                                    final Cache<String, String> nfsUsageCache, final PathGenerator pathGenerator,
                                    final FileEventManager fileEventManager, final TransferDecorator transferDecorator,
-                                   final Executor executor )
+                                   final ExecutorService executor )
     {
         this();
         this.plCacheProvider = plCacheProvider;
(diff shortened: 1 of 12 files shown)


caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
```diff
@@ -316,71 +316,113 @@ public boolean isFile( ConcreteResource resource )
...
             throws IOException
     {
-        //use "this" as lock is heavy, should think about use the transfer for the resource as the lock for each thread
+        final String pathKey = getKeyForResource( resource );
+
+        // This lock is used to control the the local resource can be opened successfully finally when local resource missing
+        // but NFS not, which means will do a NFS->local copy.
+        final Object copyLock = new Object();
+        // A flag to mark if the local resource can be open now or need to wait for the copy thread completes its work
+        final AtomicBoolean canStreamOpen = new AtomicBoolean( false );
```
jdcasey: Dumb idea maybe, but could you use this as the lock object instead of creating a separate Object? Then, I guess you'd just notifyAll() / wait() for this instead

caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
```diff
@@ -316,71 +316,113 @@ public boolean isFile( ConcreteResource resource )
...
+                    IOUtils.copy( nfsIn, localOut );
+                    logger.debug( "NFS copy to local cache done." );
+                }
+                catch ( NotSupportedException | SystemException | IOException e )
+                {
+                    if ( e instanceof IOException )
+                    {
+                        final String errorMsg =
+                                String.format( "[galley] got i/o error when doing the NFS->Local copy for resource %s",
+                                               resource.toString() );
```
jdcasey: Just use resource here, forget the toString() call. If resource is null, it'll render "null". Otherwise, it will only call toString() on the object if the log call proceeds (if the log level isn't disabled). Calling it preemptively like this incurs the string rendering cost no matter what.

caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
```diff
@@ -316,71 +316,113 @@ public boolean isFile( ConcreteResource resource )
...
+                    {
+                        canStreamOpen.set( true );
+                        copyLock.notifyAll();
+                    }
+                }
+            }
+        };
+        // This lock is used to control the concurrent operations on the resource, like concurrent delete and read/write.
+        // Use "this" as lock is heavy, should think about use the transfer for the resource as the lock for each thread
         synchronized ( getTransfer( resource ) )
```
jdcasey: I wonder if we need the whole thing (starting the copy thread and waiting to open an InputStream for the user) inside this synchronized block. It seems like synchronizing on the Transfer instance like that would avoid the case where two users race to initiate the copy from NFS into the local cache, and wind up stepping on one another (or blocking on competing writes)...

caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
```diff
@@ -316,71 +316,113 @@ public boolean isFile( ConcreteResource resource )
     public InputStream openInputStream( final ConcreteResource resource )
             throws IOException
     {
-        //use "this" as lock is heavy, should think about use the transfer for the resource as the lock for each thread
+        final String pathKey = getKeyForResource( resource );
+
+        // This lock is used to control the the local resource can be opened successfully finally when local resource missing
```
jdcasey: What if the file already exists in the local cache?

caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
```diff
@@ -534,12 +586,13 @@ public boolean delete( ConcreteResource resource )
         final String pathKey = getKeyForPath( nfsFile.getCanonicalPath() );
         synchronized ( getTransfer( resource ) )
         {
+            boolean localDeleted = false;
```
jdcasey: Would it make sense to use a while() ... wait() loop here if the file is locked?

jdcasey:
Hi Gang,

This looks pretty good overall. I'm still not convinced we've solved the locking issues, especially where we're copying files out of NFS (which requires initiating a write thread...something that we should try to de-duplicate in case two users try it at the same time). Also, I want to make sure we use the locally cached file if it's already there (for reads)...maybe I just missed that because it was in a previous commit.

I think this is enough to merge for now, and we can clean up the remainders as we go through testing a bit more.

PR #133, opened 2016-09-06
FIX: add common bmunit tests for NOS-544

yma88:
@jdcasey 
After sync and verified with gli, the original parts that I've took on NOS-544 has all coverd by NOS-556, and which code seems more considerate, so I add some common bmunit test cases for FastLocalCacheProvider, some cases are coming from CacheProviderTCK tests which could handle in byteman ways, and some are coming from what gang's single-node use cases summary. and I've extract above as AbstractFastLocalCacheBMUnitTest to extend.


jdcasey:
add to whitelist

caches/infinispan/src/test/java/org/commonjava/maven/galley/cache/infinispan/AbstractFastLocalCacheBMUnitTest.java
```diff
@@ -0,0 +1,320 @@
...
+        {
+            this.res = res;
+            this.latch = latch;
+        }
+
+        @Override
+        public void run()
+        {
+            provider.lockWrite( res );
+            provider.waitForWriteUnlock( res );
```
jdcasey: shouldn't we be trying to lockWrite() then unlockWrite()? Waiting for write unlock seems like something you do if it's already locked (or in case it's already locked). This part of the API is a little broken, admittedly...

caches/infinispan/src/test/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProviderTest.java
```diff
@@ -15,138 +15,129 @@
...
+    {
+        final Location loc = new SimpleLocation( "http://foo.com" );
+        final String path = "my/path.txt";
+        final ConcreteResource res = new ConcreteResource( loc, path );
 
-    private final TransferDecorator decorator = new TestTransferDecorator();
+        CountDownLatch latch = new CountDownLatch( 2 );
 
-    private Cache<String, String> cache = CACHE_MANAGER.getCache( NFSOwnerCacheProducer.CACHE_NAME );
+        new Thread( new WriteLockThread( res, latch ) ).start();
```
jdcasey: I think I'd pull the thread init + start out into a method in the super-class, like this:

```
protected void start(Runnable r){
    new Thread(r).start();
}
```

This gives us a nice central way to make adjustments to the way all threads are started, like if we want to use an Executor, or use a thread naming scheme or something. We can also wrap the Runnable in something else for timing, if we want to...and it removes a little boilerplate code at the same time.

It's not that important, but it may save time later if we need to do that (later we'll probably have even more test cases).

caches/infinispan/src/test/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProviderTest.java
```diff
@@ -15,138 +15,129 @@
...
-        assertNull( result );
+        final String content = "This is a test";
+        final Location loc = new SimpleLocation( "http://foo.com" );
+        final String fname = "/path/to/my/file.txt";
+        final Location loc2 = new SimpleLocation( "http://bar.com" );
+
+        CountDownLatch latch = new CountDownLatch( 2 );
+
+        new Thread( new CopyFileThread( loc, loc2, fname, latch ) ).start();
+        new Thread( new WriteFileThread( content, loc, fname, latch ) ).start();
```
jdcasey: From the test method name, I thought there was a read thread in here too...? It might make sense to have a very small javadoc that gives the initial state and expected final state (or result) of the test.

caches/infinispan/src/test/resources/bmunit/common/WriteCopyAndReadNewFile.btm
```diff
@@ -0,0 +1,21 @@
...
+DO
+    debug("<<<copyFile is waiting for service enter writeFile");
+    waitFor("service writeFile");
+    debug("<<<proceed with copyFile")
+ENDRULE
+
+RULE service writeFile
+CLASS FastLocalCacheProvider
+METHOD unlockWrite
+AT EXIT
```
jdcasey: Using Partyline (or in this case, something that uses Partyline), it should be possible to read from a file *while* it's being written. So, you might be able to use AT ENTER (if that's the term) instead of EXIT. I'm not sure, it might require watching a method deeper in the call stack (JoinableFile CTOR for example, or even EXIT of JoinableFileManager.openOutputStream()), but we should find a way to test that we can read from a file at the same time we're writing to it.

caches/infinispan/src/test/resources/bmunit/common/WriteDeleteAndVerifyNonExistence.btm
```diff
@@ -0,0 +1,21 @@
+RULE service deleteFile
```
jdcasey: So, this set of rules forces the write to complete before delete can proceed, is that true? What happens if you trigger both rules AT ENTRY, such that it's trying to delete and write simultaneously?

jdcasey:
I'm merging this manually to fix the conflicts I created when I merged ligangty's PR.

PR #135, opened 2016-09-08
NOS-503 Create RoutingCacheProvider

ligangty:


caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/routes/RoutingCacheProviderWrapper.java
```diff
@@ -0,0 +1,241 @@
...
+
+    public RoutingCacheProviderWrapper( final RouteSelector selector, final CacheProvider disposable,
+                                        final CacheProvider safe )
+    {
+        this.disposable = disposable;
+        this.safe = safe;
+        this.selector = selector;
+    }
+
+    public CacheProvider getRoutedProvider( ConcreteResource resource )
```
jdcasey: I think this might leak some implementation details to the outside world that they don't need to know about. Is it possible to reduce its scope a bit, maybe to protected?

caches/infinispan/src/test/java/org/commonjava/maven/galley/cache/routes/RoutingCacheProviderWrapperTest.java
```diff
@@ -0,0 +1,124 @@
...
+                                                                               partylineFac ).create( pathgen,
+                                                                                                      decorator,
+                                                                                                      events );
+        final CacheProvider partyline = partylineFac.create( pathgen, decorator, events );
+        final CacheProvider fastLocal = fastLocalFac.create( pathgen, decorator, events );
+
+        final String fname = "/path/to/my/file.txt";
+        Location loc = new SimpleLocation( "remote:foo/com" );
+        ConcreteResource resource = new ConcreteResource( loc, fname );
+        CacheProvider get = router.getRoutedProvider( resource );
```
jdcasey: Okay, I see how this is used now. For testing, I'd prefer if you reduce scope to protected in the main implementation, then provide a TestRoutingCacheProvider class inside this test class...that can override the getRoutedProvider() method with public scope, and delegate to it, like this:

```
private static final class TestProvider extends RoutingCacheProviderWrapper{
    public CacheProvider getRoutedProvider( ConcreteResource resource ){
        return super.getRoutedProvider(resource);
    }
}
```

If this is internal and final, it can't be abused. Also, it allows the selection method to remain internal to the routing provider, to prevent a leaky abstraction.

jdcasey:
@ligangty Overall, this looks very good. I just had a couple of nit-picky points, but nothing that should really pose a major problem. I'll be happy to merge if we can limit the scope on the getRoutedProvider() method.

ligangty:
@jdcasey, thanks for comments. Will close this and give another PR with fix.

jdcasey:
@ligangty For future reference, you can just add a new commit to the branch and it will update your open PR. You don't have to close and reopen.

PR #137, opened 2016-09-09
NOS-503 Create RoutingCacheProvider

ligangty:
Hi, @jdcasey 

The scope of the getRoutedProvider method has been limited to protected in this PR

And for that TestRoutingCacheProvider, I think it is not needed because the getRoutedProvider is still accessible in the test because it is in the same package with class for testing. I think this is a trick we usually used for the unit-testing. :)


jdcasey:
Cool. I'll merge as soon as the CI build comes back

jdcasey:
retest this please

PR #138, opened 2016-09-12
FIX: perfect some tiny tweak for NOS-544

yma88:


jdcasey:
Looks like there's some deadlock biting us here. I think it could be in partyline, but not sure.

jdcasey:
retest this please

yma88:
Hi, @jdcasey 
I mvn build in local successfully and check the output log and no dead lock found, could you point it out? or I could change the code accordingly.
While, I re build again, the mvn build process sometime would be blocked by Running org.commonjava.maven.galley.cache.infinispan.FastLocalCacheProviderConcurrentTest, very occasionally, I'm not sure if it's the issue you said.

jdcasey:
@yma88 I think the problem is the deadlock that @ligangty has been wrestling with.

jdcasey:
retest this please

jdcasey:
@yma88 I just merged a new PR for partyline which I'm hoping will help with this build.

PR #140, opened 2016-09-19
Some fix for FastLocalCacheProvider

ligangty:
1. For infinispan cache cofnig, changes the jmx domain config to galley
  2. Changes some nfs root dir check rules, just check if it is blank
  3. Will create the nfs root dir in getNFSDetachedFile if it does not exist
  4. Wraps the init method into constructors of FastLocalCacheProvider to let it always run after construction
  5. Adds two debug log in RoutingCacheProviderWrapper
  6. Changes the DualOutputStreamWrapper.close method to let it always close the wrapped streams even if Exception
  4. Changes unit-test due to these changes

### DIFF
--- a/caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/NFSOwnerCacheProducer.java
+++ b/caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/NFSOwnerCacheProducer.java
@@ -55,7 +55,7 @@ public class NFSOwnerCacheProducer
     protected EmbeddedCacheManager getCacheMgr()
     {
         final EmbeddedCacheManager cacheManager = new DefaultCacheManager(
-                new GlobalConfigurationBuilder().globalJmxStatistics().jmxDomain( "org.commonjava" ).build() );
+                new GlobalConfigurationBuilder().globalJmxStatistics().jmxDomain( "org.commonjava.maven.galley" ).build() );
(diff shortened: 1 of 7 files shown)


caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/NFSOwnerCacheProducer.java
```diff
@@ -55,7 +55,7 @@ public Configuration getCacheCfg()
     protected EmbeddedCacheManager getCacheMgr()
     {
         final EmbeddedCacheManager cacheManager = new DefaultCacheManager(
-                new GlobalConfigurationBuilder().globalJmxStatistics().jmxDomain( "org.commonjava" ).build() );
+                new GlobalConfigurationBuilder().globalJmxStatistics().jmxDomain( "org.commonjava.maven.galley" ).build() );
```
jdcasey: What is the JMX domain used for here?

ligangty: If we do not specify a jmx domain here, it will always report an exception said something like  "ISPN000034: There's already a JMX MBean instance type=CacheManager,name="CDIExtensionDefaultCacheManager" already registered under 'org.infinispan' JMX domain.".

jdcasey:
ah, so that's how you work around it. Cool.

PR #142, opened 2016-09-20
Wraps NFSCacheProducer in infinispan package only to make it produce …

ligangty:
Found that when uses the NFSCacheProducer as CDI producer for other galley user
which has its own ISPN container(like Indy), it always reports "ISPN000034" error with
jmx domain problems. I doubt it is caused by the jmx conflict when use CDI injection(not
sure), so removes the producer feature and wraps it as a default ISPN cache usage in
FastLocalCacheProvider. Any galley users can use its own ISPN cache by providing it
in FastLocalCacheProviderFactory, or leave null by use the default one.


caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
```diff
@@ -107,47 +107,29 @@
...
+    protected FastLocalCacheProvider( final PartyLineCacheProvider plCacheProvider,
                                    final Cache<String, String> nfsUsageCache, final PathGenerator pathGenerator,
                                    final FileEventManager fileEventManager, final TransferDecorator transferDecorator,
                                    final ExecutorService executor, final String nfsBaseDir )
     {
-
         this.plCacheProvider = plCacheProvider;
-        this.nfsOwnerCache = nfsUsageCache;
+        if ( nfsUsageCache != null )
+        {
```
jdcasey: I don't really think we should support the null case here. Let's force the user to supply something, and if they want to call the NFSOwnerCacheProducer themselves, they can. We can leave the class in place but just not use it automatically.

caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProviderFactory.java
```diff
@@ -51,6 +51,10 @@ public FastLocalCacheProviderFactory( File cacheDir, File nfsDir, Cache<String,
         this.executor = executor;
     }
 
+    public FastLocalCacheProviderFactory(File cacheDir, File nfsDir, ExecutorService executor){
+        this(cacheDir, nfsDir, null, executor);
```
jdcasey: Is there a reason to support this? I suppose it makes it easier to embed, but my concern is that it may lead to poor performance characteristics if we create a cache instance on-the-fly.

caches/infinispan/src/test/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProviderTest.java
```diff
@@ -161,7 +161,7 @@ public void testConstructorWitNFSSysPath()
     {
         System.setProperty( FastLocalCacheProvider.NFS_BASE_DIR_KEY, temp.newFolder().getCanonicalPath() );
         new FastLocalCacheProvider( new PartyLineCacheProvider( temp.newFolder(), pathgen, events, decorator ), cache,
-                                    pathgen, events, decorator, executor );
+                                    pathgen, events, decorator, executor, null );
```
jdcasey: I guess you'd have to call the NFSCacheOwnerProducer here instead if we eliminate null support in the provider itself.

ligangty: This null is for nfsBaseDir to test missing nfsBaseDir scenario. It is not the cache param. The cache param is using the Producer to get cache here.

ligangty:
Hi, @jdcasey 

  New change added based on your comments. Thanks.

caches/infinispan/src/test/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProviderTest.java
jdcasey: cool, guess I misread that

PR #143, opened 2016-09-21
FIX: NOS-549, Implied repos should not be processed from a profile

yma88:
FIX: NOS-549, Implied repos should not be processed from a profile

### DIFF
--- a/maven/src/main/java/org/commonjava/maven/galley/maven/model/view/MavenPomView.java
+++ b/maven/src/main/java/org/commonjava/maven/galley/maven/model/view/MavenPomView.java
@@ -656,8 +656,7 @@ public class MavenPomView
     }
 
     /**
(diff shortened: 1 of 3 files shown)


maven/src/main/java/org/commonjava/maven/galley/maven/model/view/MavenPomView.java
```diff
@@ -668,7 +672,13 @@ public ParentView getParent()
                 continue;
             }
 
-            result.add( new RepositoryView( node.getPomView(), node.getElement(), node.getOriginInfo() ) );
+            RepositoryView repoDeclaration =
```
jdcasey: Actually, can you put this in a new method, called something like:

```
public List<RepositoryView> getNonProfileRepositories()
```

getAllRepositories implies that it will return any repo declaration from anywhere in the pom.xml...and this change alters that logic.

PR #146, opened 2016-09-27
FIX: NOS-550, Implied repos should replace property placeholders

yma88:


maven/src/main/java/org/commonjava/maven/galley/maven/model/view/MavenPomView.java
```diff
@@ -1047,8 +1047,8 @@ public String resolveExpressions( final String value, final String... activeProf
             if ( ssi == null )
             {
                 ssi = new StringSearchInterpolator();
-                ssi.addValueSource( new MavenPomViewVS( this, activeProfileIds ) );
             }
+            ssi.addValueSource( new MavenPomViewVS( this, activeProfileIds ) );
```
jdcasey: I guess the first time we try to retrieve a value from the pom and it resolves expressions, this will be set, and the list of active profiles will be entered. This will affect future expression resolving calls, regardless of whether the profileId lists for those calls are different.

What this means is that we're keeping state at the MavenPomView level that doesn't belong there.

Consider what will happen if we add the list of profileIds when resolving an expression in a RepositoryView that is contained in a profile. In this case, we will start looking in the profile for expression values, which is what we want for that case. BUT, what happens if the next call is to retrieve the version of a dependency that's *not* contained in a profile, and that version has an expression in it? In this case, the expression resolver will have the opposite problem...it might resolve a property value from the profile when we shouldn't be looking at that profile.

The solution here is to get rid of the StringSearchInterpolator field, and instead construct one in this method. Just doing this will fail though, because if you have two properties like this:
```
<field-one>${field-two}</field-one>
<field-two>${field-one}</field-two>
```

and some dependency that uses the version `${field-one}`, then the method call will recurse infinitely. StringSearchInterpolator is smart enough to prevent this, but only if we do one of two things:

1. use the same SSI
2. maintain a single instance of RecursionInterceptor, and pass that into all calls

The recursion can happen via MavenPomViewVS, which calls resolveXPathExpression(..) which calls resolveExpressions(..) (this method). 

I think we need to make this method setup a new RecursionInterceptor (really `PrefixAwareRecursionInterceptor`), then immediately call a new copy of this method that declares the RecursionInterceptor as a parameter. This will pass the interceptor into the MavenPomViewVS ctor. Then, we add this interceptor as a parameter in resolveXPathExpression() as well, and call the new form of resolveExpressions from that, passing the interceptor around the whole time.

This will allow us to declare a local instance of StringSearchInterpolator in resolveExpressions (the new one variant), and call interpolate with the interceptor as the second parameter.

maven/src/test/java/org/commonjava/maven/galley/maven/model/view/MavenPomViewTest.java
```diff
@@ -387,4 +387,28 @@ public void repositoriesWithNonProfiles()
         assertFalse( repoId.contains( "profile.repository" ) );
         assertEquals( 1, repoId.size() );
     }
+
+    @Test
+    public void repositoriesWithUrlPropertyInProfile()
```
jdcasey: we need a variant of this test that retrieves a dependency from the main POM (*not* from a profile), and gets a version that contains an expression in it...and does that *after* looking at the url in the profile repository. The property should have one value in the main pom, and a different value in the profile. The assertion should be that the dependency version has the value from the main pom. So, the sequence will be:

1. Read the POM -> MavenPomView
2. Get the repository that's declared in the profile -> RepositoryView
3. Get the URL from that repository -> assert it is resolved using values from the profile
4. Get the dependency declared in the main body of the POM (not from a profile)
5. Get the version of the dependency, which is declared with an expression referencing a property that is defined one way in the main pom.xml, and another way in the profile
6. Assert the dependency version uses the value from the main pom.xml

This should show the cross-pollution of profile information I was talking about above, in your changes for MavenPomView.

maven/src/test/resources/pom-processor/repositories-expression-in-a-profile/test-pom-1.0.pom.xml
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jdcasey: This is a good test to check, but the one I mentioned above might involve redeclaring a property value for repo.one.url and using it in all of the repository declarations. For repositories in profiles where the value is redeclared, it should resolve to a value from the profile. For repositories declared in the main pom.xml, it should use the value from the main pom.xml.

Another interesting twist we should check is to leave the property declaration out of the second profile, and use the same expression for the repository url. This should resolve to the value from the main pom.xml.

yma88:
@jdcasey
I've push another commit for this, pls check.

maven/src/main/java/org/commonjava/maven/galley/maven/model/view/MavenPomView.java
```diff
@@ -1034,30 +1038,30 @@ protected boolean containsExpression( final String value )
...
+    public String resolveExpressions( final String value, final RecursionInterceptor ri,
+                                      final String... activeProfileIds )
+    {
+        StringSearchInterpolator ssi = new StringSearchInterpolator();
+        ssi.addValueSource( new MavenPomViewVS( this, ri, activeProfileIds ) );
         try
         {
-            String result = ssi.interpolate( value );
-            //            logger.info( "Resolved '{}' to '{}'", value, result );
+            String result = ssi.interpolate( value, ri );
```
jdcasey: I think you need a null check on `ri` here, and construct a RecursionInterceptor if it is null.

maven/src/main/java/org/commonjava/maven/galley/maven/model/view/MavenPomView.java
```diff
@@ -1034,30 +1038,30 @@ protected boolean containsExpression( final String value )
...
 
-        synchronized ( this )
-        {
-            if ( ssi == null )
-            {
-                ssi = new StringSearchInterpolator();
-            }
-            ssi.addValueSource( new MavenPomViewVS( this, activeProfileIds ) );
-        }
+        List<String> prefixs = new ArrayList<>();
```
jdcasey: Since you have to do the null check in `resolveExpressions(String, RecursionInterceptor, String...)` anyway, you could probably move this down to that method, and call with a null `ri` here...

maven/src/test/java/org/commonjava/maven/galley/maven/model/view/MavenPomViewTest.java
```diff
@@ -395,20 +395,28 @@ public void repositoriesWithUrlPropertyInProfile()
...
+        assertThat( rvs.get( 0 ).getName(), equalTo( "repo.one" ) );
+        assertThat( rvs.get( 0 ).getUrl(), equalTo( "http://repo.one.repository" ) );
+        assertThat( rvs.get( 1 ).getName(), equalTo( "test.oracle.repo" ) );
+        assertThat( rvs.get( 1 ).getUrl(), equalTo( "http://test.oracle.repository" ) );
+        assertThat( rvs.get( 2 ).getName(), equalTo( "test.second.oracle.repo" ) );
+        assertThat( rvs.get( 2 ).getUrl(), equalTo( "http://another.test.two.oracle.repository" ) );
+    }
+
+    @Test
+    public void resolveExpressionsBothInDepAndProfile()
```
jdcasey: Do you know if this failed before your changes to RecursionInterceptor usage? It's a little different from how I would have implemented this test, but I think it should verify the things we care about. I was wondering if it failed before, just in case my code reading missed something...

yma88: Hi, @jdcasey

Sorry I'm just back from the long holiday, and since limited in vpn, so didn't check your email in time.

I was pushing another commit for this, pls check.

And about your concern on test method: resolveExpressionsBothInDepAndProfile, before using RecursionInterceptor, some part of the tests shoud be passed, and for some test cases like redeclare the properties both in main and profile, it seems not passed IIRC, it will be failed into a wrong sistuation when using ssi to do some iteration, not all the correct active-profiles' profile-properties will be retrieved for current repository expression.

Anyway, after using the RecursionInterceptor, I try to cover all the test cases in the pom file including your comments' cases, and they all pass and run in the correct way. So I update this test and add all the test cases that I could cover into the test methods as you see.

maven/src/main/java/org/commonjava/maven/galley/maven/model/view/MavenPomView.java
```diff
@@ -119,7 +121,8 @@ public MavenPomView( final ProjectVersionRef ref, final List<DocRef<ProjectVersi
         return new HashSet<String>( resolveXPathExpressionToAggregatedList( "/profiles/profile/id/text()", false, -1 ) );
     }
 
-    public String resolveMavenExpression( final String expression, final String... activeProfileIds )
+    public String resolveMavenExpression( final String expression, final RecursionInterceptor ri,
```
jdcasey: Sorry to be picky, but this is public API, and we're going to need to maintain the old method signature. I'd suggest making the old method signature into a new method that delegates to this one with a null ri, then let it be instantiated down the call stack in resolveExpressions(..). This way we preserve backward compatibility while also fixing the property substitution.

jdcasey:
retest this please

PR #149, opened 2016-10-06
Adding CacheInstance as a mechanism for preventing cache operations when the system starts shutting down

jdcasey:


jdcasey:
retest this please

PR #151, opened 2016-10-24
NOS-653 [Galley] Change SpecialPathManager with new splitted maven support functions

ligangty:


ligangty:
@jdcasey I've used your second suggestion with a new SpecialPathSet interface to do Instance<> CDI in this PR. Now still two questions:
1. Now in the SpecialPathManagerImpl.initPkgPathSets(), all the SpecialPathInfo in the SpecialPathSet will be added into the special paths, which means the special paths will be mixed in use for all packages. I'm really not sure if this is reasonable. (For example, maybe a maven-metadata.xml is a valid file in a python repo, but will be filtered as a metadata because the java special paths contains it).

2. Not sure if SpecialPathConstants.MavenSpecialPathSet is correctly created, and if it should be moved to indy instead of in galley. Does that make sense?

jdcasey:
@ligangty I have some ideas about this, but I'm out of time for today. I do want to assign a package type to a repository in Indy, and that will mean we will have to alter SpecialPathManager to have another variant of its methods that takes a package name. SpecialPathSet will have to supply a package name too, to map the two together and sort of scope the request for a given path (by supplying a package type for it).

I think we should designate 'maven' as the default package type, for backward compatibility using the existing methods.

This has several implications for Indy's own architecture, so I'll try to document that tomorrow.

ligangty:
@jdcasey seems this pr cannot be merged now as there are many API signature changes which will cause indy compilation error. I'll continue on the indy refactor with these changes. 
BTW, any problems on this PR, leave comments and I'll fix them. Thanks.

api/src/main/java/org/commonjava/maven/galley/model/Transfer.java
```diff
@@ -227,7 +227,7 @@ public boolean exists()
         OverriddenBooleanValue overriden = null;
         if ( decorator != null )
         {
-            overriden = decorator.decorateExists( this );
+            overriden = decorator.decorateExists( this, new EventMetadata(  ) );
```
jdcasey: At some point we will probably need a variant of exists() that accepts an EventMetadata parameter, in case the existence decorator needs to know if a path has special properties within the scope of some package type.

Not critical now.

api/src/main/java/org/commonjava/maven/galley/model/Transfer.java
```diff
@@ -249,7 +249,7 @@ public void copyFrom( final Transfer f )
         {
             if ( decorator != null )
             {
-                decorator.decorateCopyFrom( f, this );
+                decorator.decorateCopyFrom( f, this, new EventMetadata(  ) );
```
jdcasey: another place where a variant of `copyFrom()` may need event metadata eventually

api/src/main/java/org/commonjava/maven/galley/model/Transfer.java
```diff
@@ -311,7 +311,7 @@ public boolean delete( final boolean fireEvents, final EventMetadata eventMetada
         String[] listing = provider.list( resource );
         if ( decorator != null )
         {
-            listing = decorator.decorateListing( this, listing );
+            listing = decorator.decorateListing( this, listing, new EventMetadata(  ) );
```
jdcasey: and again, `list()` will need a variant accepting EventMetadata eventually

core/src/main/java/org/commonjava/maven/galley/io/SpecialPathManagerImpl.java
```diff
@@ -15,61 +15,129 @@
...
     {
-        specialPaths.add( pathInfo );
+        //FIXME: do we need a flag of "isStandard" to add into stdSpecialPaths?
+        registerSpecialPathInfo( pathInfo, PKG_TYPE_MAVEN );
+    }
+
+    @Override
+    public void registerSpecialPathInfo( SpecialPathInfo pathInfo, final String pkgType )
+    {
+        for ( Map.Entry<String, SpecialPathSet> entry : pkgtypes.entrySet() )
```
jdcasey: Can you just do `pkgtypes.get(pkgType)` here?

core/src/main/java/org/commonjava/maven/galley/io/SpecialPathManagerImpl.java
```diff
@@ -15,61 +15,129 @@
...
+        pkgtypes.put( MVN_SP_PATH_SET.getPackageType(), MVN_SP_PATH_SET );
     }
 
     @Override
-    public synchronized void registerSpecialPathInfo( SpecialPathInfo pathInfo )
+    public void registerSpecialPathInfo( SpecialPathInfo pathInfo )
     {
-        specialPaths.add( pathInfo );
+        //FIXME: do we need a flag of "isStandard" to add into stdSpecialPaths?
+        registerSpecialPathInfo( pathInfo, PKG_TYPE_MAVEN );
```
jdcasey: This should be registered to the "standard" special paths, IMO.

core/src/main/java/org/commonjava/maven/galley/io/SpecialPathManagerImpl.java
```diff
@@ -15,61 +15,129 @@
...
+            }
+        }
     }
 
     @Override
     public synchronized void deregisterSpecialPathInfo( SpecialPathInfo pathInfo )
     {
-        specialPaths.remove( pathInfo );
+        //FIXME: do we need a flag of "isStandard" to remove from stdSpecialPaths?
+        deregisterSpecialPathInfo( pathInfo, PKG_TYPE_MAVEN );
```
jdcasey: Again, I think this should evaluate against the standard special path info, not the stuff for the maven package

core/src/main/java/org/commonjava/maven/galley/io/SpecialPathManagerImpl.java
```diff
@@ -15,61 +15,129 @@
...
+    }
+
+    @Override
+    public void deregisterSpecialPathInfo( SpecialPathInfo pathInfo, String pkgType )
+    {
+        for ( Map.Entry<String, SpecialPathSet> entry : pkgtypes.entrySet() )
```
jdcasey: `pkgtypes.get(pkgType)`...

core/src/main/java/org/commonjava/maven/galley/io/SpecialPathManagerImpl.java
```diff
@@ -15,61 +15,129 @@
...
+            logger.warn(
+                    "[Galley] The package types already contains the path set for this package type {}, will override it",
+                    pathSet.getPackageType() );
+        }
+
+        pkgtypes.put( pathSet.getPackageType(), pathSet );
+    }
+
+    @Override
+    public void deregesterSpecialPathSet( SpecialPathSet pathSet )
```
jdcasey: IMO we should return the removed SpecialPathSet. It's similar to Map semantics, and if people don't need the return value they can just discard it.

maven/src/main/java/org/commonjava/maven/galley/maven/ArtifactManager.java
```diff
@@ -80,10 +80,10 @@ Transfer store( Location location, ArtifactRef ref, InputStream stream )
     Transfer store( Location location, ArtifactRef ref, InputStream stream, EventMetadata eventMetadata )
         throws TransferException;
 
-    boolean publish( Location location, ArtifactRef ref, InputStream stream, long length )
+    boolean publish( Location location, ArtifactRef ref, InputStream stream, long length, EventMetadata eventMetadata )
```
jdcasey: This one is a little bit weird. It's a subsection of galley that is maven-specific, so I think it would be okay to leave the EventMetadata parameter out (if we were just using it for package type)...but then again, we use EventMetadata for other things now, so it's reasonable to have it here. I guess let's leave it like you have it.

api/src/main/java/org/commonjava/maven/galley/TransferManager.java
```diff
@@ -75,16 +75,16 @@ boolean delete( ConcreteResource resource )
     boolean delete( ConcreteResource resource , EventMetadata eventMetadata  )
         throws TransferException;
 
-    boolean publish( ConcreteResource resource, InputStream stream, long length )
+    boolean publish( ConcreteResource resource, InputStream stream, long length, EventMetadata metadata )
```
jdcasey: Ideally these should have the old method signatures preserved, where they call the new one with a `new EventMetadata()` to default over to Maven package type.

I remember, when I introduced EventMetadata in the first place, that adding these original methods back in made it harder to find consistency problems in the calls I made from Indy into TransferManager...we can leave it until later to restore backward compat if you like...as long as it happens before the next galley release, I think we should be okay.

ligangty:
@jdcasey Thanks for your comments. I've changed/added all api sigs back for backward-compatibility in new commit. 
BTW, not sure why, when I did compatibility testing on indy with these new galley changes, I got this test failure very occasionally(not always):

Expected: <false>
     but: was <true>
	at org.commonjava.indy.mem.data.ConcurrencyTest.deadlockOnGroupContains(ConcurrencyTest.java:85)
Caused by: java.lang.AssertionError: 

Expected: <false>
     but: was <true>

Not sure if this is caused by galley changes.

core/src/main/java/org/commonjava/maven/galley/io/SpecialPathManagerImpl.java
```diff
@@ -62,41 +62,27 @@ public void initPkgPathSets()
...
     public void deregisterSpecialPathInfo( SpecialPathInfo pathInfo, String pkgType )
     {
-        for ( Map.Entry<String, SpecialPathSet> entry : pkgtypes.entrySet() )
-        {
-            if ( pkgType.equals( entry.getKey() ) )
-            {
-                entry.getValue().deregisterSpecialPathInfo( pathInfo );
-            }
-        }
+        pkgtypes.get(pkgType).deregisterSpecialPathInfo( pathInfo );
```
jdcasey: oops, looks like I didn't follow my own recommendations when I wrote these methods...ugh :(

jdcasey:
@ligangty I'll take a look at that assertion...it's really not part of the actual test. That test is mostly about triggering deadlock (or verifying that it doesn't trigger)...so the assertion is a bit of a distraction. Still, it's strange that the expectation and actual values are empty, since it's trying to verify a boolean. Maybe it has to do with the fact that it's executing in a thread.

core/src/main/java/org/commonjava/maven/galley/io/SpecialPathManagerImpl.java
ligangty: Sometimes when I gives these types of code, it will lead me to a inertial thinking to loop and fetch firstly... or maybe some unknown mess in my mind makes me think in confusion.... :(

PR #163, opened 2017-03-07
Write http-metadata.json filee up front instead of after the transfer completes

ruhan1:
@jdcasey Fixed according to your comment with a small twist. I make changes in HttpDownload and HttpExistence rather than their super abstract class because writeMetadata() takes parameters which are conveniently available in derived classes. This is the minimal code change. Further, it is not as good to distinguish GET/HEAD request from others in the super class as to distinguish those via natural polymorphism in derived classes. A ftest is added to Indy.


jdcasey:
@ruhan1 IMO we need to integrate the writeMetadata() method into the executeHttp() call in a finally block, and with a condition that it only runs on HEAD or GET. This would allow us to capture metadata on failed requests too, which is critical for responding consistently to the user after a failed request is cached in the NFC.

ruhan1:
I didn't realize that the failed request needs metadata too. I updated the PR. please review again.

PR #174, opened 2017-04-12
[MASTER] Adds some tracing log for Transfer delete operation

ligangty:


jdcasey:
minimal risk, IMO not worth waiting to merge.

PR #181, opened 2017-04-25
Not cache for certain path when SpecialPathInfo.isCachable() return false

ruhan1:


core/pom.xml
```diff
@@ -34,6 +34,10 @@
       <artifactId>galley-api</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.commonjava.maven.galley</groupId>
```
jdcasey: What's this new dependency used for? If possible, I'd like to reduce its scope to `test` or maybe eliminate it from the core module build, since it shouldn't really be necessary for a user to have a dependency on the httpclient transport in order to use the core functions...in principle, they should be able to use galley with a ssh, ftp, zip, or even just file transports, without http. I know it's unlikely, but still...

jdcasey: I'm going to go ahead and merge this PR but I'd like to follow up on this before we're completely done.

ruhan1: Well, this is only for using constant HttpExchangeMetadata.FILE_EXTENSION - ".http-metadata.json". I was not sure when I added the dep for such a minor convenience. Where else can we put this constant (maybe in api? or we just use the literal ".http-metadata.json")

jdcasey:
@ruhan1 I think I might suggest moving the constant and renaming to something like HTTP_METADATA_EXT. You could put it in SpecialPathConstants or something

ruhan1:
@jdcasey , added it to SecialPathConstants. Removed the dep from core. Keep the original constant in HttpExchangeMetadata (because many places use it) and assign the new constant to it. A new PR pushed, pls review.

PR #184, opened 2017-05-03
Enable a seprate logger to track delete operation logs

ligangty:


jdcasey:
Closing this in favor of #186, which uses a more specific logger name for content deletion, makes it a constant we can reference elsewhere for similar content deletion logs, and provides a stack trace if trace logging is enabled (so we can see what's calling the delete).

PR #185, opened 2017-05-03
[0.13.x]Enable a seprate logger to track delete operation logs

ligangty:


jdcasey:
Closing for two reasons:

1. I superceded it with #186 which provides a more specific logger name
2. I'd rather not port things like this back to the 0.13.x branch unless we need to...I'm hoping we can stop maintaining that branch ASAP

PR #186, opened 2017-05-03
Log content deletion to a logger specifically named for the purpose

jdcasey:
This uses a separate logger, with the 'org.commonjava' base namespace, an
intermediate namespace of 'content' (for any content action), and a
specific name of 'delete'. This will allow us to isolate these messages
from others and push them to a different log appender. We should also
be able to reuse the name for any location where content is being deleted
in Commonjava libraries / applications.

This PR supercedes @ligangty's. Thanks for doing the initial work!


jdcasey:
Original work in #184

PR #190, opened 2017-05-17
Add closed flag to resolve "Double-close" chaos

ligangty:


ligangty:
retest this please

caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
```diff
@@ -96,6 +96,12 @@
 
     private PathGenerator pathGenerator;
 
+    // This thread holder is used to add some "re-entrant" like function for the ISPN transaction lock. The ISPN lock is
+    // used for ISPN transaction but not for thread level with re-entrant, that means if we want to own this lock in one
+    // transaction for more than twice in single thread, it will block this thread. So we need this thread holder to by-pass
+    // the ISPN lock waiting when thread not changed.
+    private final ThreadLocal<Long> threadHolder = new ThreadLocal<>();
```
jdcasey: I have two concerns here:

1. What will clear this ThreadLocal? If you're acting only through the CacheProvider interface, and responding to a close() call to clear this, how do you know whether you're dealing with the close() call for a derivative (eg. checksum) file, or the original file that caused the lock to be created?

2. Can a single thread have two open files in different directories? If so, will one overwrite the ThreadLocal, and make it even harder to know when to clear it? 

Usually in cases like these, we resort to using reference counting, so that each new lock/open operation increases the lock count, and each subsequent unlock/close operation decreases it. Then, if the reference count reaches zero, clear the lock.

IIRC, this is what ReentrantLock does.

Also, to keep one open operation from stomping on another, we should probably keep separate locks for each directory...in a map.

I think this re-proves the need for a map of ReentrantLock keyed on directory path, similar to the one FileTree uses in partyline. If you use a Map<String, ReentrantLock>, then you'll still have to synchronize the lazy initialization of a new ReentrantLock, but it's a small synchronization point. 

If you used a map of ReentrantLock, I think you could also get rid of synchronizing on Transfer and just use the ReentrantLock (probably).

NOTE: If Transfer itself has synchronized methods, you might have to implement something like:

```
    <T> T CacheProvider.synchronizeAnd( transfer, ()->{ /*some logic*/});
```

Then, in FastLocalCacheProvider:

```
    public <T> T synchronizeAnd( Transfer transfer, Supplier<T> operation){
        // lookup and lock the ReentrantLock
        return operation.get();
        // finally: release lock
    }
```

In PartylineCacheProvider we might implement as:

```
    public <T> T synchronizeAnd( Transfer transfer, Supplier<T> operation){
        // NOTE: This might need to be exposed in JoinableFileManager
        return filemanager.lockAnd( transfer.getDetachedFile(), operation);
    }
```

And in FileCacheProvider, it might be simpler:

```
    public <T> T synchronizeAnd( Transfer transfer, Supplier<T> operation){
        synchronized( transfer ){
            return operation.get();
        }
    }
```

I'm not sure how this stuff would affect the cluster, though. The lock ownership is still maintained in the cache, which is clustered. The ReentrantLock would cause local threads to wait, but remote threads should wait because the ownership is still registered in the cache. What am I missing?

jdcasey: Thinking about it more, I don't think we necessarily want ReentrantLock mapped to path. Instead, maybe we just want a set of locked directories local to the thread. ThreadLocal will help us with the "local to the thread" part, but I might prefer using ThreadContext from weft for this, so it will be used across multiple associated threads if we do stream post-processing in parallel.

I'm not sure how this would impact the cluster though. It's locked at the directory level, and that lock owner is shared through the ISPN cluster, right? Other local threads trying to access files in the directory would wait on the lock owner, and threads in other parts of the cluster should do likewise, shouldn't they?

ligangty:
retest this please

ligangty:
Hi, @jdcasey I've updated the code with some change like:
* Use ThreadContext instead to hold thread id
* Add some "wait before lock" for ISPN lock to avoid duplicate lock operation(which may cause javax.transaction.NotSupportedException from ISPN)

And I'm still think the re-entrant feature you pointed out with a counter update way. Seems that it is not a fully identical case to the normal "re-entrant" for the thread. It is just used to avoid the ISPN lock blocked by more than two lock operations on the same directory "locker" in one single thread. But I can not make sure now if this is a reasonable way. I need to think about some common and uncommon cases for it, and then update with what I sort out.(Maybe some new sequence diagrams are needed)

PR #193, opened 2017-06-02
Nos 921

ligangty:
New u-tests revealed that when single thread writes two files in same
folder with abnormal sequence(open 1 -> open 2 -> close 1 -> close 2),
the ISPN transaction will throw some transaction exception for
transaction duplicates or transaction has been ended. This commit will add
a thread state counter of one single thread to identify if the
transaction should be ended correctly if there are several opened files
simultaneously without close. Current solution is if one file opened, it
will add a count to the thread count and open transaction only if there
was no transaction yet. And only when all thread count has been
removed(threadState is 0), will close the transaction. That means
whatever if these files are in the same folder(use same path lock) or not(use
different), it will all be in one single transaction in one thread.


jdcasey:
So, it looks like you're re-deriving the solution I've been working on in partyline, with some minor differences. With this, we're going to have three locking solutions here: the one you're working on now, the ISPN lock, and the partyline lock.

However, we only really need two: The partyline lock and the ISPN lock.

Why don't we manage thread access and re-entry using partyline (with a change I just made). After all, when will we need to lock ISPN that we don't also want a lock on partyline?

When we start an operation that needs locking, we can use JoinableFileManager on the local system to lock the directory, then once we have the local lock, we can acquire the ISPN lock. The group should be lockable multiple times in partyline (again, with the latest change), which will mean that you can simply unlock the directory then immediately call JoinableFileManager.isLockedByCurrentThread()...if that returns false, you know it's time to unlock the ISPN cache.

The partyline change for this is:

https://github.com/Commonjava/partyline/pull/65

On the other hand, if we want to do a clean LIFO style for the locking (partyline.lock() -> ISPN.lock() -> ISPN.unlock() -> partyline.unlock()) we can implement another method in JoinableFileManager easily which would return the lock count for the current ThreadContext. When this reaches 1, we unlock ISPN, then unlock partyline.

It just seems like managing these two locking systems should be enough. I think the solution you're working on here is so similar to partyline that we can consolidate them and wrap ISPN with it.

ligangty:
Hi, @jdcasey 
 I've replaced the thread re-entrant solution with your solution in partyline(Added a lockTimes api here: https://github.com/Commonjava/partyline/pull/66 ).
 But I found one issue here. When I run this unit test: 
https://github.com/ligangty/galley/blob/a1475ef8ed210139b61500c7fdf1e33c388ac9bb/caches/infinispan/src/test/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProviderIOTest.java#L249
It will fail with a re-submit(rollback) exception of ISPN. And after debugging, I found that for one single directory lock, the thread lock count always counts to 1 (no increase). And the problem line is here:
https://github.com/Commonjava/partyline/blob/master/src/main/java/org/commonjava/util/partyline/FileTree.java#L319
In this code block, when one thread lock a directory more than twice, seems that each time a new lockOwner will be created and updated in entryMap, which result the lockOwner.locks owner info is always new (If I understand correctly). Not sure if this is a bug.

jdcasey:
@ligangty I updated partyline's FileTree logic that you referenced, and commented the three main cases we have to handle there. See https://github.com/Commonjava/partyline/pull/67

Also, I made a minor request to rename the method for retrieving the context-specific lock count in partyline, that's on PR https://github.com/Commonjava/partyline/pull/66

jdcasey:
retest this please

caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
```diff
@@ -100,6 +101,12 @@
 
     private final JoinableFileManager fileManager = new JoinableFileManager();
 
+    // This file counter is used to solve the problem of multi file operations in on ISPN TX. Sometimes in one single ISPN TX,
+    // it may includes more than one file operations. If we don't control the ISPN lock of each operation separately and just use
+    // tx.commit/rollback, it may cause ISPN TX in some weird state. This file counter is more like a thread re-entrant feature for
+    // ISPN single TX for different files.
+    private final AtomicInteger fileCounter = new AtomicInteger( 0 );
```
jdcasey: Won't this cross-pollute between paths, such that two threads accessing different paths will use the same counter, and only one of the paths' transactions will be closed?

* Thread Context 1: open file (increment counter == 1)
* Thread Context 2: open file2 (increment same counter == 2)
* Thread Context 2: close file2 (decrement counter == 1; file2 transaction NOT closed)
* Thread Context 1: close file( decrement counter == 0; close file transaction)

Should this counter be stored in ThreadContext?

ligangty:
@jdcasey new PR submitted for your comments. The file counter is moved into ThreadContext. 
I think I still need to add some u-tests for your case there, so this PR is still not completed. Will ping you later when all's ready.

PR #196, opened 2017-06-27
NOS-946, Implement readonly protections at Transfer/CacheProvider level

yma88:


api/src/main/java/org/commonjava/maven/galley/model/SimpleLocation.java
```diff
@@ -31,15 +31,17 @@
...
 
     private final Map<String, Object> attributes = new HashMap<String, Object>();
 
     private final String name;
 
     public SimpleLocation( final String name, final String uri, final boolean allowSnapshots, final boolean allowReleases,
- final boolean allowsStoring, final boolean allowPublishing,
-                           final boolean allowDownloading )
+                           final boolean allowsStoring, final boolean allowPublishing, final boolean allowDownloading,
+                           final boolean allowDeletion )
```
ligangty: I'm not sure if any other projects are using this class, so suggest to retain former constructor and add this as new one. Just set the allowDeletion to true as default for the former one.

core/src/main/java/org/commonjava/maven/galley/internal/TransferManagerImpl.java
```diff
@@ -696,6 +697,10 @@ public boolean delete( final ConcreteResource resource, final EventMetadata even
         throws TransferException
     {
         final Transfer item = getCacheReference( resource );
+        if ( !resource.allowsDeletion() )
+        {
+            throw new TransferException( "Deletion not allowed for: {}", resource );
+        }
```
ligangty: move this to the first line of this method for optimization.

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/model/SimpleHttpLocation.java
```diff
@@ -37,20 +37,20 @@
...
-        super( name, uri, allowSnapshots, allowReleases, allowsStoring, allowPublishing, true );
+        super( name, uri, allowSnapshots, allowReleases, allowsStoring, allowPublishing, true, allowDeletion );
         this.urlInfo = new UrlInfo( uri );
         this.proxyUrlInfo = proxyUri == null ? null : new UrlInfo( proxyUri, 8080 );
     }
 
     public SimpleHttpLocation( final String name, final String uri, final boolean allowSnapshots,
                                final boolean allowReleases, final boolean allowsStoring, final boolean allowPublishing,
-                               final String proxyUri, LocationTrustType trustType )
+                               final boolean allowDeletion, final String proxyUri, LocationTrustType trustType )
```
ligangty: Same consideration for this class with SimpleLocation constructor.

PR #197, opened 2017-07-26
Add new constructor of HttpExchangeMetadata with HttpServletRequest and Response

yma88:
,for ContentController http metadata generate usage.


jdcasey:
I think I'd rather see this functionality in a utility class in Indy, to avoid adding extra dependencies that don't affect the transport itself.

PR #201, opened 2017-08-29
generate npm tarball and version metadata files from package.json metadata when publish

yma88:


core/src/main/java/org/commonjava/maven/galley/internal/TransferManagerImpl.java
```diff
@@ -626,8 +631,15 @@ public Transfer store( ConcreteResource resource , final InputStream stream , fi
         OutputStream out = null;
         try
         {
+            String content = IOUtils.toString( stream );
```
jdcasey: If the content is a 200Mbyte jar this code could be pretty bad for memory performance, since it's storing the whole thing in memory.

core/src/main/java/org/commonjava/maven/galley/internal/TransferManagerImpl.java
```diff
@@ -943,4 +955,65 @@ public boolean publish( final ConcreteResource resource, final InputStream strea
         return batch;
     }
 
+    private void generateNPMFilesFromTarget( final String content, ConcreteResource resource, final Transfer target,
```
jdcasey: We've kept all of the maven-specific pieces out of this library (they're actually in ArtifactManager, in galley-maven). I don't think NPM-specific stuff belongs in galley-core. Is this something we can handle using a TransferDecorator or maybe just in the Indy codebase?

Also, we need to allow the STORAGE_PATH attribute to be used for things that aren't NPM...so the presence of this attribute shouldn't be enough to trigger a whole bunch of assumptions about NPM.

We should be able to intercept the store / create code in the Indy NPM handler, and do something like this...though I'd still prefer to see us write the content to a temp file (NOT in NFS storage if we can avoid it), then read the parts using a stream-oriented parser...and writing the tgz content back to disk without storing it in memory. Again, if all of that is possible.

Tarballs can get pretty big. Keeping these in memory will limit how far Indy can scale when supporting NPM builds.

api/src/main/java/org/commonjava/maven/galley/model/Transfer.java
```diff
@@ -513,4 +519,14 @@ public void setResource( ConcreteResource resource )
     {
         this.resource = resource;
     }
-}
+
+    public void setGeneratedFromTarget( Set<Transfer> generatedFromTarget )
```
jdcasey: Can we handle this split in the Indy code? It seems strange to keep state about "real" Transfers that were split out of this Transfer

yma88:
@jdcasey I took whole day to move this galley npm part to indy, and now some CDI work is on there still, not finished yet.

It's definitely right to make most of the npm logic into npm module, here some of the changes such as let NPMContentHandler to deal the metadatas generation logic, I need to pick it up back to the npm module (IIRC at the beginning of npm desgin, we don't choose the npm self handler and put all the logic into ContentAcessHandler, now it's more complex that original, self handler is necessary), if so, some of the 'STORAGE_PATH' I also recommand to couple in it if possible.

Although the STORAGE_PATH is not only used for npm in future, by now the path mapping conversion is still as the npm's road, customized for npm, such as doing the path mapping when store, download, retrieve(only for hosted/group), I'm not sure they must be general for other package types.

Another important thing is big tarball and package.json metadata mid writing , I almost understand your thinking, while could you introduce some of the ways or examples to implement? I just realized for package.json merged when publish or group, it's also possible to deal two big streams, right? still need something like temp file storage and reading the parts using a stream-oriented parser... (I only test the case using small package, I really need to find a real case to verify!)
If the process above should not be happend in galley, as you said we need a intercept for npm handler, something I need to find examples too.

yma88:
close this, has moved the stuff into indy commit: https://github.com/Commonjava/indy/pull/764/commits/a1e57b8b8ec8cf6546049bfca40fb16f5283e857

PR #202, opened 2017-09-07
[NPM] avoid some package.json redirect path empty errors (visit by the listing), especially for the group

yma88:
such as for the conent listing:
if not fix this, hosted will show: jquery/
remote will show: jquery
group will show: jquery/

while if someone visits remote or group content (include a remote at least) as jquery/, will report some empty error. This will also effect the group merging result if some one visit jquery/ before jquery, this fix will prevent it from the listing visit side.

after fix this, all the types of repo will show the contents like 'jquery' (just like a file).

Acturally, we should do the listing or html page showing like other npm registries in future? they will not show any projects on the index listing, only the json of some metadatas about the regisitry, while unified metadata json between hosted and remote should be definite, I know this is not a high priority issue, just found this when doing the manual tests.

### DIFF
--- a/core/src/main/java/org/commonjava/maven/galley/internal/TransferManagerImpl.java
+++ b/core/src/main/java/org/commonjava/maven/galley/internal/TransferManagerImpl.java
@@ -313,6 +313,13 @@ public class TransferManagerImpl
                         {
                             for ( String fname : fnames )
                             {
+                                // npm will only show the project as a file path no matter it's a real file or path,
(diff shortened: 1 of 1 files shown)


jdcasey:
@yma88 IMO we need to rethink how listings work. This looks fine, but you make a good point about providing a similar experience with other repos like npmjs. We might need to add some mechanism at a higher level (Indy) to cope with the expectations of different package repo types.

PR #203, opened 2017-09-07
[NOS-928] Content-Length mimatch check [NOT ready]

bxf12315:
@jdcasey 
Please reivew it.

### DIFF
--- a/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
+++ b/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
@@ -15,19 +15,9 @@
  */
 package org.commonjava.maven.galley.transport.htcli.internal;
 
-import static org.apache.commons.io.IOUtils.closeQuietly;
-import static org.apache.commons.io.IOUtils.copy;
-import static org.commonjava.maven.galley.spi.cache.CacheProvider.STORAGE_PATH;
-
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.util.Map;
-
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
-import org.apache.http.HttpStatus;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.util.EntityUtils;
 import org.commonjava.maven.galley.TransferContentException;
@@ -41,8 +31,14 @@ import org.commonjava.maven.galley.transport.htcli.model.HttpLocation;
 import org.commonjava.maven.galley.transport.htcli.util.HttpUtil;
 import org.commonjava.maven.galley.util.ResourceUtils;
 
-import java.util.Collection;
-import java.util.Collections;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Map;
+
+import static org.apache.commons.io.IOUtils.closeQuietly;
+import static org.apache.commons.io.IOUtils.copy;
+import static org.commonjava.maven.galley.spi.cache.CacheProvider.STORAGE_PATH;
 
 public final class HttpDownload
     extends AbstractHttpJob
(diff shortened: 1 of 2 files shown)


transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
```diff
@@ -137,7 +142,25 @@ private void writeTarget()
...
+                * we will delete the file and return a 502 Gateway Error status to indicate
+                * the upstream server is having problems.
+                */
+                Header[] headers = response.getHeaders( "Content-Length" );
+                Collection<Integer> successStatuses = Collections.singleton( HttpStatus.SC_OK );
+                if ( headers[0] !=null && !headers[0].getValue().equals( "" ))
+                {
+                    long contentLength = Long.parseLong( headers[0].getValue().toString() );
+                    long targetLength = target.length();
+                    if ( contentLength != targetLength && successStatuses.contains( response.getStatusLine().getStatusCode() ) )
```
jdcasey: What if the content length header isn't provided? I think we need a null check here too.

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
```diff
@@ -137,7 +142,25 @@ private void writeTarget()
             try
             {
                 final HttpEntity entity = response.getEntity();
-
+                /*
+                * If the server responded with 200 but the content length doesn't match,
+                * we will delete the file and return a 502 Gateway Error status to indicate
```
jdcasey: I think it would help for future reference to change this comment to something like "...we will delete the file and return a null Transfer as if the file wasn't found, since the content is invalid." You know, just because we moved away from the idea of returning a 502 for this...

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
jdcasey: I *think* we only call `writeTarget()` when we have a success status...in other words, that we've already checked for 200 status before we get here. If that's true, then we shouldn't need to check it again, right?

ligangty: Or if we want to check, why not directly HttpStatus.SC_OK==response.getStatusLine().getStatusCode()?

bxf12315: @jdcasey yes you are right. we was checked the status in https://github.com/Commonjava/galley/blob/8eb579da6e06cb16e9abce0638437b5ea8781df1/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/AbstractHttpJob.java#L112
@ligangty  Now HttpStatus.SC_OK just 200, maybe in future we will add other success status

ligangty: After a further check of the code, I've some concerns which I don't know if it is right. As this download job is executed in TransferManager.retrieve(), which will check if the the target transfer exists, so that means in this job the target transfer should not exist, and if this is true, why should we check the target transfer length with response header "content-length"?

bxf12315:
@jdcasey My last commit can be work.
please review it again.

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
```diff
@@ -179,6 +158,32 @@ private void writeTarget()
...
+                {
+                    if ( contentLength != target.length() )
+                    {
+                        try
+                        {
+                            target.delete();
+                        }
+                        catch ( final Exception e )
+                        {
+                            throw new TransferException(
```
jdcasey: You can't throw an exception from a finally clause. If the finally executes after an exception in the try section, then an exception is thrown from the finally section, one of the two will be lost. **This is VERY dangerous.**

If you need to execute this regardless of try / catch, then store the exception from the catch into a variable, then put this logic below the finally section, and make a decision about what to throw...documenting your decision and logging both errors no matter what.

**HOWEVER** I'd rather see this in the try section, and then in the catch clause just call `target.delete()`. If there's an exception that we're throwing, the associated Transfer is automatically invalid IMO.

bxf12315: You are right. It is very dangerous , throw exception in finally section.
I remove the code section because of we ready length of target need close the outputstream.
So, I can close it in try block and check content length, then close it again in finally block.
Is it right ?

jdcasey:
@ligangty This looks okay to me, but if you find something I missed, please bring it up (maybe as a new issue for our tracking purposes) and we can address it. I'm going to merge this now in the interest of maybe getting an Indy 1.2.0 release today or tomorrow.

jdcasey:
@bxf12315 I reverted this commit. Please re-submit a new PR once you have the tests working.

PR #210, opened 2017-09-27
[0.15.x] improve equality check for pending transfers, better pending sync

jdcasey:


core/src/main/java/org/commonjava/maven/galley/internal/xfer/DownloadHandler.java
```diff
@@ -109,15 +109,15 @@ private Transfer joinOrStart( final ConcreteResource resource, final Transfer ta
...
-        if ( target.exists() )
-        {
-            return target;
-        }
-
         Future<DownloadJob> future;
         synchronized ( pending )
         {
+            // if the target file already exists, skip joining.
+            if ( target.exists() )
```
ruhan1: It looks great. I think this solves the race condition you mentioned in jira comment. Btw, it would be easy to understand if we name the "pending" in the sync to something like DOWNLOAD_MUTEX, and make the pending itself a normal map. I will submit a PR for it.

PR #211, opened 2017-09-28
[NOS 928] [Not Ready] avoid cache for TransferContentException

bxf12315:
@jdcasey 
It is just my guess approach.
I'm not sure it is work.
I will try to test it.

### DIFF
--- a/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
+++ b/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
@@ -154,6 +154,7 @@ public final class HttpDownload
                 ConcreteResource resource = target.getResource();
                 try
                 {
+                    logger.debug( "Failed to write to local proxy store :{} so delete target:{}",e,target.getPath() );
                     target.delete();
                 }
                 catch ( IOException e1 )
--- a/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
+++ b/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
@@ -158,9 +158,14 @@ public final class HttpDownload
                 }
                 catch ( IOException e1 )
                 {
+                    logger.error( "Failed to delete target file: {}\nOriginal URL: {}. Reason: {}", e1,
+                                  target, url, e1.getMessage() );
+
                     throw new TransferException( "Failed to delete target file: {}\nOriginal URL: {}. Reason: {}", e1,
                                                  target, url, e1.getMessage() );
                 }
+                logger.error( "Failed to write to local proxy store: {}\nOriginal URL: {}. Reason: {}", e, target, url,
+                              e.getMessage() );
                 throw new TransferContentException( resource,
                                                     "Failed to write to local proxy store: {}\nOriginal URL: {}. Reason: {}",
                                                     e, target, url, e.getMessage() );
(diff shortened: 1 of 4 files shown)


transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
```diff
@@ -147,8 +148,23 @@ private void writeTarget()
             }
             catch ( final IOException e )
             {
-                throw new TransferException( "Failed to write to local proxy store: {}\nOriginal URL: {}. Reason: {}",
-                                             e, target, url, e.getMessage() );
+                ConcreteResource resource = null;
+                if ( target.exists() )
```
jdcasey: I think we can avoid checking if it exists and just delete it. But first, you'll have to use IOUtils.closeQuietly(out) to ensure the file can be deleted.

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
```diff
@@ -147,8 +148,23 @@ private void writeTarget()
...
+                {
+                    try
+                    {
+                        resource = target.getResource();
+                        target.delete();
+                    }
+                    catch ( IOException e1 )
+                    {
+                        throw new TransferException(
```
jdcasey: Log this to the error output too, and make sure you log the original IOException to error too so we don't lose it. I agree this will be the higher priority error and needs to be thrown instead of TransferContentException, but we need to ensure the original error information is in the log. In fact, I'd also include the original error.getMessage() in the TransferException too.

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
bxf12315: Yes.
These code in catch block,In finally block we have IOUtils.closeQuietly(out) , I think finally always 
execute firstly. Is it good enough?

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
bxf12315: ok

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
jdcasey: This is the opposite of my understanding of how finally works. Can you please verify this before we rely on that effect? 

In any case, IMO it improves clarity of the code to have the extra `closeQuietly()` call in the catch block, and it will not have a severe impact on performance.

bxf12315: Sorry, My fault , finally block just before throw exception and break the method.
Honestly ,I never write logic in catch block, so I'm misunderstanding the basic concept. 
I will using it.

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
```diff
@@ -148,22 +148,21 @@ private void writeTarget()
...
-                        resource = target.getResource();
-                        target.delete();
-                    }
-                    catch ( IOException e1 )
-                    {
-                        throw new TransferException(
-                                        "Failed to write to local proxy store: {}\nOriginal URL: {}. Reason: {}", e1,
-                                        target, url, e1.getMessage() );
-                    }
+                    throw new TransferException( "Failed to delete target file: {}\nOriginal URL: {}. Reason: {}", e1,
```
jdcasey: If you throw this TransferException without logging the IOException `e`, we lose the original cause of the problem. That will add days or weeks to any production problem that might happen in this code, and require us to fix & redeploy to production before we can even debug the problem.

Please log (to ERROR log level) `e1` and throw TransferException with the original IOException as the cause here.

jdcasey:
I folded the changes here into #216 and merged it there with some additional fixes / testing.

PR #219, opened 2017-11-06
Improve copy-on-read sync for FastLocalCacheProvider

ruhan1:


ruhan1:
retest this please

caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
```diff
@@ -352,6 +361,10 @@ else if ( e instanceof InterruptedException )
             {
                 while ( !canStreamOpen.get() )
                 {
+                    if ( copyExceOccurs.get() )
+                    {
+                        return null;
+                    }
                     try
                     {
                         copyLock.wait();
```
jdcasey: This wait needs to have a timeout on it so the while condition can be checked continuously. I'd set it around 100ms.

ruhan1: I think this does not need a timeout. The main thread waits here until it is waked up by the copy thread (when either read-ready or copy-error). The while just makes sure it go to right path after it is waked up. We don't need a continuous check here.

ruhan1:
Two tests failed. I am debugging them.

PR #220, opened 2017-11-14
Some new tests for big file concurrent R/W to test performance

ligangty:
I've added some tests both for FLCProvider and PartyLineCachProvider, to test big file io performance for both provider, but seems it is not good enough, even some suspicious dead lock there. I just ignored that 2 tests, and try to find what's wrong there.


ligangty:
@jdcasey This test PR is finished, if you think it is ok, I'll merge it.

PR #233, opened 2018-01-31
Implement http-metadata generation for file uploading

ligangty:


transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/UploadMetadataGenTransferDecorator.java
```diff
@@ -0,0 +1,138 @@
...
+            }
+        }
+
+        final HttpExchangeMetadata metadata = new HttpExchangeMetadataFromRequestHeader( requestHeaders );
+        OutputStream out = null;
+        try
+        {
+            final Transfer finalMeta = metaTxfr;
+            out = metaTxfr.openOutputStream( TransferOperation.GENERATE, false );
+            logger.debug( "Writing HTTP exchange metadata:\n\n{}\n\n", new Object()
```
jdcasey: Need to revise to trace() for the final version.

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/UploadMetadataGenTransferDecorator.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    private void writeMetadata( final Transfer target, final ObjectMapper mapper, final Map<String, List<String>> requestHeaders )
+    {
+        Transfer metaTxfr = target.getSiblingMeta( HttpExchangeMetadata.FILE_EXTENSION );
+        if ( metaTxfr == null )
+        {
+            if ( target.isDirectory() )
+            {
+                logger.debug( "DIRECTORY. Using HTTP exchange metadata file INSIDE directory called: {}", HttpExchangeMetadata.FILE_EXTENSION );
```
jdcasey: Does this only happen if someone is trying to write to a directory as if it was a file? I'm tempted to say we should log that as an error, since it's insane...but that's something the I guess the CacheProvider should throw the error for?

I'd revise this one to trace()

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/UploadMetadataGenTransferDecorator.java
```diff
@@ -0,0 +1,138 @@
...
+                metaTxfr = target.getChild( HttpExchangeMetadata.FILE_EXTENSION );
+            }
+            else
+            {
+                logger.debug( "SKIP: Cannot retrieve HTTP exchange metadata Transfer instance for: {}", target );
```
jdcasey: Probably trace() here too

PR #237, opened 2018-03-09
TransferDecorator implementations not using package-type variants of SpecialPathManager methods

jdcasey:
I noticed that TransferDecorators are using the old SpecialPathManager methods that
default to using the 'maven' package type. This is not always appropriate.

I've corrected this by using EventMetadata.getPackageType() in these decorators.


jdcasey:
retest this please

PR #238, opened 2018-03-13
Update to Maven 3.5.0 for plugin defaults

rnc:
@jdcasey Utilised https://github.com/apache/maven/blob/maven-3.5.0/maven-core/src/main/resources/META-INF/plexus/default-bindings.xml and https://github.com/apache/maven/blob/master/maven-core/src/main/resources/META-INF/plexus/default-bindings.xml to locate default versions.

### DIFF
--- a/maven/src/main/java/org/commonjava/maven/galley/maven/GalleyMavenBuilder.java
+++ b/maven/src/main/java/org/commonjava/maven/galley/maven/GalleyMavenBuilder.java
@@ -21,7 +21,6 @@ import org.commonjava.maven.galley.TransferManager;
 import org.commonjava.maven.galley.cache.CacheProviderFactory;
 import org.commonjava.maven.galley.maven.internal.ArtifactManagerImpl;
 import org.commonjava.maven.galley.maven.internal.ArtifactMetadataManagerImpl;
-import org.commonjava.maven.galley.maven.internal.defaults.StandardMaven304PluginDefaults;
 import org.commonjava.maven.galley.maven.internal.defaults.StandardMaven350PluginDefaults;
 import org.commonjava.maven.galley.maven.internal.defaults.StandardMavenPluginImplications;
 import org.commonjava.maven.galley.maven.internal.type.StandardTypeMapper;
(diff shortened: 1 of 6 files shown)


testing/maven/src/main/java/org/commonjava/maven/galley/testing/maven/GalleyMavenFixture.java
```diff
@@ -592,7 +593,7 @@ public GalleyMavenFixture withPomReader( final MavenPomReader pomReader )
         return this;
     }
 
-    public GalleyMavenFixture withPluginDefaults( final StandardMaven304PluginDefaults pluginDefaults )
+    public GalleyMavenFixture withPluginDefaults( final StandardMaven350PluginDefaults pluginDefaults )
```
rnc: There is already a `withPluginDefaults ( final MavenPluginDefaults pluginDefaults )` so perhaps this could be removed. Note there is also a matching function in CartographerCoreBuilder I believe.

jdcasey: :+1:

maven/src/main/java/org/commonjava/maven/galley/maven/GalleyMavenBuilder.java
```diff
@@ -199,7 +200,7 @@ public MavenPluginDefaults getPluginDefaults()
         return pluginDefaults;
     }
 
-    public GalleyMavenBuilder withPluginDefaults( final StandardMaven304PluginDefaults pluginDefaults )
+    public GalleyMavenBuilder withPluginDefaults( final StandardMaven350PluginDefaults pluginDefaults )
```
jdcasey: Looks like I made a mistake here, it should be generalized to MavenPluginDefaults.

testing/maven/src/main/java/org/commonjava/maven/galley/testing/maven/GalleyMavenFixture.java
jdcasey: Agreed, MavenPluginDefaults is the right parameter type.

PR #239, opened 2018-03-16
Add local file cache for expiration deletion in FLCP

ligangty:
As we will make the local cache of artifacts expiration automatically periodically, here uses a new ISPN cache to record the local file resource, and then trigger the deletion by ISPN cache expiration mechanism.


bxf12315:
retest this

bxf12315:
retest this

PR #241, opened 2018-04-24
It appears that FLCP is getting stuck while retrieving/init'ing Transfer instances

jdcasey:
The transfers cache in FLCP has been de-syncrhonized in favor of ReentrantLock. The
problem is the map of ReentrantLocks is keyed by Transfer instances, and we don't
have one of those yet. So, we need to put the synchronized keywords back in for
getTransfer(..) and clearTransferCache().

### DIFF
--- a/caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
+++ b/caches/infinispan/src/main/java/org/commonjava/maven/galley/cache/infinispan/FastLocalCacheProvider.java
@@ -915,14 +915,20 @@ public class FastLocalCacheProvider
     }
 
     @Override
-    public Transfer getTransfer( ConcreteResource resource )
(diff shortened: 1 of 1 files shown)


ligangty:
So that means the ContextSensitiveWeakHashMap is not thread-safe? Maybe we can implement a thread-safe version from ConcurrentWeakHashMap from ispn which is lock-free.

ligangty:
Have a quick review of the ConcurrentWeakHashMap and seems it is not lock-free, which used ReentrantLock too. So lock is still needed.

jdcasey:
@ligangty I'm not entirely sure if it's a concurrency issue or just delay from working through a huge backlog of garbage collectable Transfer instances. I don't have anything but a stack trace that's suggestive. I'm not sure I understand your comment about ConcurrentWeakHashMap...does that mean that my change is necessary, or...?

PR #245, opened 2018-05-17
Support metric to measure HTTP transport

ruhan1:


transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
```diff
@@ -55,27 +59,75 @@
...
     public DownloadJob call()
+    {
+        if ( metricConfig == null || !metricConfig.isEnabled() || metricRegistry == null )
+        {
+            return doCall();
+        }
+
+        logger.trace( "Download metric enabled, location: {}", location );
+
+        Timer repoTimer = null;
```
jdcasey: Can we create a generic second timer for all http downloads and use that here too? That way, we have a way to sort of aggregate / average all upstream calls for everything including the ones we don't single out, and can to 99th-percentile graphing of that time to look for a repo that might be dragging down performance.

Also, can we get call counts as well as timing info?

ruhan1: A globalTimer is a few lines below for all downloads. :)
According to https://stackoverflow.com/questions/30987757/dropwizard-metrics-meters-vs-timers, a timer also provides "a count of the number of events".

PR #250, opened 2018-08-15
switch to using scoring logic to find a preferred vs. 'good-enough' IPv4 address

jdcasey:


jdcasey:
We're not using this for now, but while we have the error report for the error while running in an OpenShift container with a particular configuration, I wanted to take care of it.

thescouser89:
I'm unsure if the first IP address will be the local IP address though if the method call  `isSiteLocalAddress` fails

PR #255, opened 2018-10-19
Snapshot filter not working when allowSnapshot disabled

ligangty:


api/src/main/java/org/commonjava/maven/galley/io/TransferDecoratorPipeline.java
```diff
@@ -136,7 +136,11 @@ public void decorateDelete( final Transfer transfer, final EventMetadata metadat
         String[] result = listing;
         for ( TransferDecorator decorator : decorators )
         {
-            result = decorator.decorateListing( transfer, listing, metadata );
+            String[] decorated = decorator.decorateListing( transfer, listing, metadata );
+            // Only decorated listing changed need to override the result.
+            if(!Arrays.equals( listing, decorated )){
```
ruhan1: The if statement seems not necessary. if the decorated is equals to listing, the assignment is still valid. 
btw, during the for loop, it use an immutable listing. I wonder should here use result instead of listing? i.e., 
String[] decorated = decorator.decorateListing( transfer, result, metadata );
in case the result changed.

ligangty: This if statement is really necessary, because all other Decorator is using default decorateListing from AbstractTransferDecorator, which just simply returned the original listing. The only one did the changed work is ContentsFilterTransferDecorator, but its result is override by other ones, which actually did not change the result. I don't think this is a correct action for this decorateListing method, and from the local testing work, it proved my thoughts: the ContentsFilterDecorator.decorateListing result is override by other ones which did not change listing, which is not the correct case for issue NOS-1434.
So we should only consider the decorated result which actually give changes to the listing result here. And for advance, maybe we still need to think more about this decorateListing logic: is the simple override logic is correct here for the content listing logic?

ruhan1: Can we use 'result' in the param instead of 'listing'? i.e.,
result = decorator.decorateListing( transfer, result, metadata );
After ContentsFilterTransferDecorator changed the 'result', the rest of these decorators use the updated 'result' as the param and return it.

ligangty: Maybe this decorateListing is not so accurate from semantic level in this TransferDecoratorPipeline, because if there are more than one Decorator, the following ones may override the result of the previous ones, which are not correct. (Even we use the previous result to do the following logic).
But here I agree to use the result instead but not original listing, because it will not break the logic. So will change.

ligangty: Maybe this decorateListing is not so accurate from semantic level in this TransferDecoratorPipeline, because if there are more than one Decorator, the following ones may override the result of the previous ones, which are not correct. (Even we use the previous result to do the following logic).
But here I agree to use the result instead but not original listing, because it will not break the logic. So will change.

ligangty: Maybe this decorateListing is not so accurate from semantic level in this TransferDecoratorPipeline, because if there are more than one Decorator, the following ones may override the result of the previous ones, which are not correct. (Even we use the previous result to do the following logic).
But here I agree to use the result instead but not original listing, because it will not break the logic. So will change.

ligangty: Maybe this decorateListing is not so accurate from semantic level in this TransferDecoratorPipeline, because if there are more than one Decorator, the following ones may override the result of the previous ones, which are not correct. (Even we use the previous result to do the following logic).
But here I agree to use the result instead but not original listing, because it will not break the logic. So will change.

ligangty: Maybe this decorateListing is not so accurate from semantic level in this TransferDecoratorPipeline, because if there are more than one Decorator, the following ones may override the result of the previous ones, which are not correct. (Even we use the previous result to do the following logic).
But here I agree to use the result instead but not original listing, because it will not break the logic. So will change.

ligangty: Maybe this decorateListing is not so accurate from semantic level in this TransferDecoratorPipeline, because if there are more than one Decorator, the following ones may override the result of the previous ones, which are not correct. (Even we use the previous result to do the following logic).
But here I agree to use the result instead but not original listing, because it will not break the logic. So will change.

ligangty: Maybe this decorateListing is not so accurate from semantic level in this TransferDecoratorPipeline, because if there are more than one Decorator, the following ones may override the result of the previous ones, which are not correct. (Even we use the previous result to do the following logic).
But here I agree to use the result instead but not original listing, because it will not break the logic. So will change.

ligangty: Maybe this decorateListing is not so accurate from semantic level in this TransferDecoratorPipeline, because if there are more than one Decorator, the following ones may override the result of the previous ones, which are not correct. (Even we use the previous result to do the following logic).
But here I agree to use the result instead but not original listing, because it will not break the logic. So will change.

ligangty: Maybe this decorateListing is not so accurate from semantic level in this TransferDecoratorPipeline, because if there are more than one Decorator, the following ones may override the result of the previous ones, which are not correct. (Even we use the previous result to do the following logic).
But here I agree to use the result instead but not original listing, because it will not break the logic. So will change.

ligangty: So I'm not sure if this decorateListing is correct from semantic level: the following decorators' result can still override the previous ones. 
But in current context, I agree to use result instead as it will not break the ContentsFilterDecorator's result. So will change.

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/ContentsFilteringTransferDecorator.java
```diff
@@ -129,6 +137,26 @@ public OutputStream decorateWrite( final OutputStream stream, final Transfer tra
                             result.add( element );
                         }
                     }
+
+                }
+            }
+            else
+            {
```
pkocandr: This doesn't make sense to filer. Because if you have path like "org/apache" there cannot be any snapshot (you don't have the version folder yet), that's why there was no "else" branch in this condition.

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/ContentsFilteringTransferDecorator.java
```diff
@@ -63,7 +65,9 @@ public OverriddenBooleanValue decorateExists( final Transfer transfer )
                 final String path = transfer.getPath();
                 // pattern for "groupId path/(artifactId)/(version)/(filename)"
                 // where the filename starts with artifactId-version and is followed by - or .
-                final Pattern pattern = Pattern.compile( ".*/([^/]+)/([^/]+)/(\\1-\\2[-.][^/]+)$" );
+//                final Pattern pattern = Pattern.compile( ".*/([^/]+)/([^/]+)/(\\1-\\2[-.][^/]+)$" );
+                // NOS-1434 all files with snapshot version(in snapshot folder) should be ignored if !allowsSnapshots
+                final Pattern pattern = Pattern.compile( ".*/([^/]+)/([^/]+)/(.[^/]+)$" );
```
pkocandr: Why is this pattern changed. Before it was designed to match only artifact, i.e. files matching pattern "<artifactId>-<version>" followed with a classifier or a file extension. This will match also any metadata and files not even on a GAV directory level that we don't really want to filter out. Or?

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/ContentsFilteringTransferDecorator.java
pkocandr: This doesn't make sense to filer. Because if you have path like "org/apache" there cannot be any snapshot (you don't have the version folder yet), that's why there was no "else" branch in this condition.

ligangty: In this case we want to filter the snapshot folder itself. For example, for the /commons-codec/commons-codec/, it will have both snapshot and normal release folder like 1.1/, 1.1-SNAPSHOT/, so we want to filer this 1.1-SNAPSHOT/ out of the listing.

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/ContentsFilteringTransferDecorator.java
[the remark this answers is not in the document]
ligangty: I remember that in the jira we want to filter out all files even metadata in the snapshot folder here, so this is why I changed pattern.

ligangty:
I'll split this pr to two. One is just for api signature fixing, another one will contain the enhance code from my part.

PR #257, opened 2018-10-23
Add attribute content-type into HttpExchangeMetadata

sswguo:
Adding the attribute content-type here that can be used to get the result from upstream.


ligangty:
retest this please

PR #259, opened 2018-10-25
Snapshot filter not working when allowSnapshot disabled

ligangty:


ligangty:
The unit test in this pr tells why I think these code change is needed. @pkocandr @jdcasey if you think the case in the test is not reasonable, just ignore this pr and I'll close it.

transports/httpclient/src/test/java/org/commonjava/maven/galley/transport/htcli/ContentFilteringTransferDecoratorTest.java
```diff
@@ -0,0 +1,188 @@
...
+    };
+
+    @Test
+    public void snapshotNotExistsWhenSnapshotsNotAllowed()
+            throws Exception
+    {
+        final String fname = "/commons-codec/commons-codec/11-SNAPSHOT/maven-metadata.xml.md5";
+
+        final String content = "kljsjdlfkjsdlkj123j13=20=s0dfjklxjkj";
+        fixture.getServer().expect( "GET", fixture.formatUrl( fname ), new ExpectationHandler()
```
jdcasey: There's a variant of this method that just accepts a string body and response status code, so you don't have to implement a handler. Just FYI

jdcasey:
@pkocandr We're getting more serious about directory browsing, and I'm concerned that leaving in SNAPSHOT directories will result in confusion. Additionally, there are cases where we use directory listings to generate maven-metadata.xml files where there are none...in this case, if we don't filter out the SNAPSHOT directories, they could show up in the list of versions.

Will this patch actively hurt anything that you know of?

PR #262, opened 2018-11-01
Fix NPE when decorateWrite for snapshot metadata filtering

ligangty:
When wrapped running of this MetadataFilteringOutputStream, the buffer will be null which caused NPE. So should assign a new instance each time.


ligangty:
@pkocandr I think this is the root cause for the snapshot metadata not filtering problem.

PR #266, opened 2018-11-20
Delete empty folder when do listing to avoid confusion

ruhan1:


ruhan1:
I tested it by 1. start a local indy; 2.issue two curls to create two folders; 3. list and can see both; 4. remove * from one storage folder; 5. list again and only see the existing one; 6. the folder itself was deleted from storage.
P.S., the curls:
curl http://localhost:8080/api/content/maven/remote/central/ae/teletronics/toolbox/cache/0.8/cache-0.8.pom
curl http://localhost:8080/api/content/maven/remote/central/ae/teletronics/toolbox/cache/0.9/cache-0.9.pom

ruhan1:
Sounds risky. I can think of another case - what if one creates a folder but before uploading any files the other delete it? Maybe we just ignore it but not delete it.

PR #267, opened 2018-11-21
Fix transfer filtering of existence

ligangty:
For the first time of access of resource, the transfer is null, which
means it's not both file and directory. So the transfer.isFile will not
work. According to this, we should use !transfer.isDirectory to filter
it out


pkocandr:
I'm a bit concerned that it is not like you describe it. If transfer was null, it would throw NPE. But you probably mean that the path does not exist, so the transfer points at a non-existing location. The problem with this change is if somebody sends an exists request on a directory. Then this condition will return true even though the target path still does not exist.

PR #269, opened 2019-01-29
[Do not merge] Just test download behavior

ruhan1:


ruhan1:
This is related to httpclient async download. We might need it when updating galley code to support async. Keep it here for a while.

PR #272, opened 2019-02-25
Remove transferCache from PartyLineCacheProvider

ruhan1:


jdcasey:
Maybe I was looking at old code previously, but didn't we have logic to delete expired content in the getTransfer() method?

ruhan1:
@jdcasey I don't think so. The field is private. The only thing to clean it is clearTransferCache and I did not find anywhere to call it. 
This is based on latest Galley code.

jdcasey:
Things must have changed. I looked at it Friday and had planned to add a mark/sweep routine to deal with the delete logic I was looking at. I guess it must've been old code.

jdcasey:
@ruhan1 would you mind upgrading Indy with this and submitting a PR with this and the new Weft and Partyline, to see if it'll build? If it won't, feel free to back out the Weft/Partyline dependency updates in Indy and I'll take a look at integrating that stuff separately.

ruhan1:
@jdcasey I think it is better not update partyline/weft ATM. Adding a seeping timer to clean the per-path lock might not the best way. I am thinking to clean it during the final block once we unlock it. If no one else is using the lock, we can just remove it. I suggest we consider those fixes later.

PR #282, opened 2019-07-11
Move ContentsFilteringTransferDecorator to Indy and update TransferPipeline

ruhan1:
This is for supporting NPM metadata masking. Briefly, we need more than one transfer decorators to handle different tasks. We did this via TransferDecoratorPipeline which itself is a TransferDecorator. This confuse things a bit. I make the TransferDecoratorManager to align all decorators. I inject the decorators via Instance&lt;&gt; but Indy still uses the old way to construct it (ie. new xxx()).


jdcasey:
I don't think anything else is particularly interested in filtering out repository metadata like this, but we can look at putting together some package-specific decorator modules eventually if we need it.

PR #283, opened 2019-08-02
Use PKG_RE_DOWNLOAD to control some metadata redownload if missed

ligangty:


jdcasey:
I think a better solution will address this in the PathGenerator, specifically the one Indy supplies. See https://github.com/Commonjava/indy/pull/1331

ligangty:
Use pathCalculator & pathgenerator way. Close this.

PR #292, opened 2019-09-18
Improve metrics for transfer decorator, attempt to improve latency

jdcasey:
This PR just attempts to reduce latency that might be due to lock collisions when writing http-metadata, and also attempts to improve the metrics reported for different transfer decorators / checksum generators. It doesn't really change much but just tries to improve visibility. I've tested all of these changes lightly in the staging environment.

### DIFF
--- a/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/UploadMetadataGenTransferDecorator.java
+++ b/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/UploadMetadataGenTransferDecorator.java
@@ -31,6 +31,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.enterprise.inject.Alternative;
 import javax.inject.Named;
+import java.io.FilterOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.List;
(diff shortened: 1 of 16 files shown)


transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/UploadMetadataGenTransferDecorator.java
```diff
@@ -48,96 +52,146 @@
...
                 metadata.get( STORE_HTTP_HEADERS ) != null && metadata.get( STORE_HTTP_HEADERS ) instanceof Map;
 
-        // We only care about non-metadata artifact uploading for http-metadata generation
         if ( isUpload && !isMetadata && hasRequestHeaders )
         {
-            @SuppressWarnings( "unchecked" ) Map<String, List<String>> storeHttpHeaders =
-                    (Map<String, List<String>>) metadata.get( STORE_HTTP_HEADERS );
-            writeMetadata( transfer, new ObjectMapper(), storeHttpHeaders );
+            // we defer writing http-metadata.json until the main stream closes, to avoid trouble with directory-level write locks.
+            return new HttpMetadataWrapperOutputStream( stream, transfer, metadata );
```
ligangty: Here seems the original stream is used. Currently is ok as AbstractTransferDecorator.decorateWrite directly return this original stream which is used, but I'm not sure if we will refactor it in future to add some other decorated logic. So maybe we should also pass the super.decorateWrite logic to this HttpMetadataWrapperOutputStream by some ways?

jdcasey:
@ligangty I found the problem with metric naming last night, as I was debugging those. I also found some potential for NPEs, which I'm fixing now. I'll work on these items, thanks!

PR #295, opened 2019-09-27
Remove physical file when there is IOException

ruhan1:
It is for NOS-2014 - If a caller opens a new OutputStream to write a file, and then there is an IOException while writing the file, the generated physical file should be deleted/marked for deletion immediately

### DIFF
--- a/caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/core/FileBasedPhysicalStore.java
+++ b/caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/core/FileBasedPhysicalStore.java
@@ -51,4 +51,11 @@ public class FileBasedPhysicalStore implements PhysicalStore
         return new FileInputStream( new File( baseDir, storageFile ) );
     }
 
+    @Override
+    public boolean delete( FileInfo fileInfo )
+    {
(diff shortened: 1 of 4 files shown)


caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/core/FileBasedPhysicalStore.java
```diff
@@ -51,4 +51,11 @@ public InputStream getInputStream( String storageFile ) throws IOException
         return new FileInputStream( new File( baseDir, storageFile ) );
     }
 
+    @Override
+    public boolean delete( FileInfo fileInfo )
+    {
+        File f = new File( baseDir, fileInfo.getFileStorage() );
+        return f.delete();
```
sswguo: Or worth to use Files.deleteIfExists(..) ? As the official doc said: 

> Note that the Files class defines the delete method to throw an IOException when a file cannot be deleted. This is useful for error reporting and to diagnose why a file cannot be deleted.

ruhan1: Sounds good. Updated.

PR #299, opened 2019-10-09
Check the duration before delete the files

sswguo:


sswguo:
> The config.getGCIntervalInMinutes was meant the interval to run Gc, usually 1 or half hours. I think we need some config called "GCGracePeriodInHours" which is by default 24 hours or so.

Yeah, sure.

PR #300, opened 2019-10-10
Use HashSet for ReverseMap.paths (JPA can handle it)

ruhan1:


caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/model/ReverseMap.java
```diff
@@ -15,13 +17,13 @@
     private ReverseKey reverseKey;
 
     @Column
-    private String paths;
+    private HashSet<String> paths;
```
sswguo: I found this annotation:  @ElementCollection 
( When the Map value is an embeddable class or basic type, use the @ElementCollection annotation. )
Do we need it ?

ruhan1: Using this one will create another table and JPA joins these two tables when querying/updating/etc. Cassandra does not support this joining so we have to avoid it.

sswguo: ACK, i see.

PR #301, opened 2019-10-14
Support datastax Cassandra PathDB

ruhan1:
Datastax API is different from JPA. I separate JPA and Datastax implements. They share the same model interfaces. 
I also add Junit tests for Cassandra with an embedded server.


caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/util/CassandraPathDBUtils.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class CassandraPathDBUtils
+{
+    public static final String PROP_CASSANDRA_HOST = "cassandra_host";
+
+    public static final String PROP_CASSANDRA_PORT = "cassandra_port";
+
+    public static final String PROP_CASSANDRA_KEYSPACE = "cassandra_keyspace";
+
+    public static String getSchema_CREATE_KEYSPACE( String keyspace )
```
ligangty: Actually from java specification, it is not good practice to use "_" in method name. Maybe getSchemaCreateKeyspace is better.

caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/util/CassandraPathDBUtils.java
```diff
@@ -0,0 +1,49 @@
...
+    {
+        return "CREATE KEYSPACE IF NOT EXISTS " + keyspace
+                        + " WITH REPLICATION = {'class':'SimpleStrategy', 'replication_factor':1};";
+    }
+
+    public static String getSchema_CREATE_TABLE_PATHMAP( String keyspace )
```
ligangty: See above.

caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/util/CassandraPathDBUtils.java
```diff
@@ -0,0 +1,49 @@
...
+                        + "filename varchar,"
+                        + "fileid varchar,"
+                        + "creation timestamp,"
+                        + "size int,"
+                        + "filestorage varchar,"
+                        + "PRIMARY KEY (filesystem, parentpath, filename)"
+                        + ");";
+    }
+
+    public static String getSchema_CREATE_TABLE_REVERSEMAP( String keyspace )
```
ligangty: See above.

caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/util/CassandraPathDBUtils.java
```diff
@@ -0,0 +1,49 @@
...
+    {
+        return "CREATE TABLE IF NOT EXISTS " + keyspace + ".reversemap ("
+                        + "fileid varchar,"
+                        + "paths set<text>,"
+                        + "PRIMARY KEY (fileid)"
+                        + ");";
+    }
+
+    public static String getSchema_CREATE_TABLE_RECLAIM( String keyspace )
```
ligangty: See above.

caches/path-mapped/pom.xml
```diff
@@ -33,13 +33,29 @@
             <groupId>javax.persistence</groupId>
             <artifactId>javax.persistence-api</artifactId>
         </dependency>
+
+        <!--Datastax Java Driver-->
+        <dependency>
+            <groupId>com.datastax.cassandra</groupId>
+            <artifactId>cassandra-driver-core</artifactId>
+            <version>3.7.2</version>
```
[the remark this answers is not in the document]
ruhan1: Sure. When moving to a new project and versions get stable I will use properties.

caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/datastax/CassandraPathDB.java
```diff
@@ -0,0 +1,348 @@
...
+
+    @Override
+    public boolean exists( String fileSystem, String path )
+    {
+        PathMap pathMap = getPathMap( fileSystem, path );
+        if ( pathMap != null )
+        {
+            return true;
+        }
+        return false;
```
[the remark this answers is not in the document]
ruhan1: Yep. That could be better. The other reason I keep that seemingly redundant code is that all of these similar methods look similar. I copy/paste the get/if lines for each of them. :)

caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/datastax/CassandraPathDB.java
```diff
@@ -0,0 +1,348 @@
...
+        PathMap pathMap = getPathMap( fileSystem, path );
+        if ( pathMap != null )
+        {
+            return pathMap.getFileStorage();
+        }
+        return null;
+    }
+
+    @Override
+    public void copy( String fromFileSystem, String fromPath, String toFileSystem, String toPath )
```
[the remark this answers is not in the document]
ruhan1: Sounds good.

caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/util/CassandraPathDBUtils.java
ruhan1: Ok.

ruhan1: Actually I don't know if I need the param... I will test it with keyspace specified somewhere else. If ok, I will roll back to simple constants.

ruhan1:
@sswguo the second commit is for gc. I updated the reclaim model, make a fack partition key, and use the deletion as a clustering column so we can query it easily.

sswguo:
> @sswguo the second commit is for gc. I updated the reclaim model, make a fack partition key, and use the deletion as a clustering column so we can query it easily.

That's cool. 👍

ruhan1:
CI failed. I merge this so others can start working on it.

PR #303, opened 2019-10-17
Add path generator for mapped-path cache provider

ligangty:
We need path generator to determine the real path from outer system, for example, npm package.json when hit the root package path


ligangty:
The main problem here is not for hashed one, but for indy one with npm handling. The main logic here for the npm case is: if a request through some path like /jquery/ or /@type/jquery/, its real path is not these two but the generated ones: /jquery/package.json or /@type/jquery/package.json. This logic is implemented through alternative pathcalculator wrapped in IndyPathGenerator, which will finally be injected in CacheProvider.

PR #304, opened 2019-10-17
Extract path handling logic to lambda based

ligangty:
And also add path generator logic to some more methods


ruhan1:
LGTM. Can you move this to https://github.com/Commonjava/path-mapped-storage . I just moved the code.

ligangty:
Maybe not needed as all change are for cache provider.

ruhan1:
+1

PR #311, opened 2019-10-22
[path-mapped] update deleteExecutor

ruhan1:
I found the Galley deleteExecutor is not used because config.isTimeoutProcessingEnabled is always false. This is fine. But if enabled, the scheduled executor only takes one item to delete (remove(0)). I change it to normal executor service.

### DIFF
--- a/caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/PathMappedCacheProvider.java
+++ b/caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/PathMappedCacheProvider.java
@@ -16,6 +16,7 @@
 package org.commonjava.maven.galley.cache.pathmapped;
 
 import org.commonjava.cdi.util.weft.SingleThreadedExecutorService;
+import org.commonjava.maven.galley.util.PathUtils;
 import org.commonjava.storage.pathmapped.core.PathMappedFileManager;
 import org.commonjava.maven.galley.io.TransferDecoratorManager;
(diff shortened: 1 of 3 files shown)


caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/PathMappedCacheProvider.java
```diff
@@ -229,47 +209,24 @@ public String getStoragePath( final ConcreteResource resource )
...
-                                                          config.getDefaultTimeoutSeconds() );
-
+        final int timeoutSeconds = getResourceTimeoutSeconds( resource );
         if ( !resource.isRoot() && config.isTimeoutProcessingEnabled() && timeoutSeconds > 0 )
         {
-            if ( isFileTimeout( txfr, timeoutSeconds ) )
-            {
-                toDelete.add( txfr );
-            }
+            deleteExecutor.submit( () -> {
```
ligangty: I'm a little curious here. Seems this may cause a not existed transfer returning if it is timeout. Am I right?

ruhan1: The isFileTimeout has a DB access so I thread it off. If the file not exist, the method will return -1.

ruhan1:
The config timeoutProcessingEnabled is not set in prev Partyline provider. I think this delete mechanism is never used.

PR #316, opened 2019-10-30
Add close() to cacheprovider so that the caller have a chance to clearup

ruhan1:
Add a close() to CacheProvider.AdminView so that the caller can do clean-up.


ruhan1:
This depend on https://github.com/Commonjava/path-mapped-storage/pull/19

PR #321, opened 2019-11-13
Avoid to store the http-metadata file when the target is not found

sswguo:
Issue: https://projects.engineering.redhat.com/browse/NOSSUP-112


jdcasey:
We shouldn't need to use them for 404 IMO

PR #322, opened 2019-11-15
Fix the NPE issue

sswguo:
Move the check into the method writeMetadata, where there is null check at the beginning for the object response.


sswguo:
@ruhan1 , I will provide another patch for filtering the specific type of files.

PR #323, opened 2019-11-18
Skip to write http metadata for some types of the files

sswguo:
Not sure if it's worth to generate a config for this and then accept it from the client(Indy).

### DIFF
--- a/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/AbstractHttpJob.java
+++ b/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/AbstractHttpJob.java
@@ -189,6 +189,12 @@ public abstract class AbstractHttpJob
             return;
         }
 
+        if ( target.getPath().endsWith( ".md5" ) || target.getPath().endsWith(".sha1") )
+        {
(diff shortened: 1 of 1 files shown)


transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/AbstractHttpJob.java
```diff
@@ -189,6 +189,12 @@ private void writeMetadata( final Transfer target, final ObjectMapper mapper )
             return;
         }
 
+        if ( target.getPath().endsWith( ".md5" ) || target.getPath().endsWith(".sha1") )
```
ruhan1: Maybe also sha256, sha512, etc. Galley a AbstractChecksumGeneratorFactory and a few impl classes. Each impl class has a fileExtension. I think maybe we shuld share something instead of hardcoding them here?

ruhan1: Maybe create an enum of checksums and save the file ext together. And use them elsewhere.

sswguo: Ah, see. Good to know that we have the impl classes, i will see how to do it later. thanks.

PR #329, opened 2020-02-03
Add the arg packageType for the method getSpecialPathInfo

sswguo:
Adding the arg pkgType to support to query npm.


ligangty:
One question here. Seems this pr only provide a signature but not the real logic for the type-specified special path getting. Do we have any attention here for differeces between maven and npm for special path getting? After all we have MavenSpecialPathSets and NPMSpecialPathSets already.

sswguo:
> One question here. Seems this pr only provide a signature but not the real logic for the type-specified special path getting. Do we have any attention here for differeces between maven and npm for special path getting? After all we have MavenSpecialPathSets and NPMSpecialPathSets already.

We actually need to get the specialPathInfos from either MavenSpecialPathSets or NPMSpecialPathSets, passing the arg pkgType is trying to identify the specific sets when I learning from the code, lines around as follows:

https://github.com/Commonjava/galley/blob/master/core/src/main/java/org/commonjava/maven/galley/io/SpecialPathManagerImpl.java#L199

PR #330, opened 2020-02-19
Clear download job in finally block

ruhan1:
If the download failed but the job is left in the pending. This causes all the future download attemps fail.

### DIFF
--- a/core/src/main/java/org/commonjava/maven/galley/internal/xfer/DownloadHandler.java
+++ b/core/src/main/java/org/commonjava/maven/galley/internal/xfer/DownloadHandler.java
@@ -152,15 +152,6 @@ public class DownloadHandler
                     logger.debug( "Waiting for download job of path: {}: {}", resource, future );
                     final DownloadJob job = future.get( waitSeconds, TimeUnit.SECONDS );
 
-                    if ( created )
-                    {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/commonjava/maven/galley/internal/xfer/DownloadHandler.java
```diff
@@ -238,6 +229,14 @@ else if ( !suppressFailures )
...
         {
             transferSizes.remove( target );
+            if ( created )
+            {
+                synchronized ( DOWNLOAD_MUTEX )
+                {
+                    logger.debug( "Removing download job of path: {}: {}", resource, future );
+                    pending.remove( target );
+                }
+            }
```
ligangty: I think the fix can work but it will bring some side effect which will make download conflict. Because there is no retry check for that pending remove, so when a big file download timeout, it will trigger a retry with current future. But pending has been removed now so if there is a new download coming in the meantime, a new future will be created to do downloading for same resource, which could bring download conflict. (or maybe a not completed target will be returned)
Maybe we need to check tries>0 before the code.

ligangty: Double checked and seems that retry is happening in inner loop. So ignore this.

PR #331, opened 2020-03-04
Fix a test util class compile err

ruhan1:
This is a test util class. But miss one interface method. Not sure why mvn did not find it...


ruhan1:
thanks @sswguo This is simple but i didn't want to make silly mistakes...

PR #333, opened 2020-03-27
Save transfer immutables

ruhan1:
Some transfer properties are immutable. We can save them because they are called more than one times per request. And in case of path-mapped storage, each is one DB access.
I tested with Indy ftests and all passed.
I picked up some ftests and calculate the db access count, this fix reduces half of path-db accesses (50%). 
Full test result is in NOS-2441.

### DIFF
--- a/api/src/main/java/org/commonjava/maven/galley/model/Transfer.java
+++ b/api/src/main/java/org/commonjava/maven/galley/model/Transfer.java
@@ -69,14 +69,35 @@ public class Transfer
         this.provider = provider;
     }
 
+    /*
+     * Some properties are immutable and we cache them in case the provider operations are expensive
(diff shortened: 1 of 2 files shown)


ruhan1:
In Indy, transfer objects life cycle should be no more than per-request. But actually it can be reused if want. This object is stateless - no fields. All it does is to calcaulte something. In theory, if we cache all transfer objects and reuse them, nothing will break. 
Actually I don't want to add fields to break the stateless. Stateless is beautiful. I did a closer look at the PathMappedCacheProvider, the findings are interesting. I will paste it in JIRA.

ruhan1:
I will drop this. It is not perfect and breaks the pure steteless.

PR #335, opened 2020-05-19
[NFC] Ignore the timeout exception when addMissing

sswguo:
I think we can ignore to put the target in NFC when it run into the TransferTimeoutException. Sometimes it failed at the first time and might success when retry that next time.


ruhan1:
LGTM

PR #340, opened 2020-06-26
Write sub array when necessary.

whitingjr:
This PR patches #339 to ensure the checksum is updated only with the intended array contents. It might be all or smaller quantity of array elements. The change caters for the smaller quantity case.

### DIFF
--- a/core/src/main/java/org/commonjava/maven/galley/io/checksum/ChecksummingOutputStream.java
+++ b/core/src/main/java/org/commonjava/maven/galley/io/checksum/ChecksummingOutputStream.java
@@ -131,17 +131,14 @@ public final class ChecksummingOutputStream
     }
 
     @Override
-    public void write (byte[] b, int off, int len)
-    throws IOException
+    public void write ( byte[] b, int off, int len )
+            throws IOException
     {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/commonjava/maven/galley/io/checksum/ChecksummingOutputStream.java
```diff
@@ -131,17 +132,19 @@ public void write( final int data )
...
             throw new IndexOutOfBoundsException();
 
         super.write( b, off, len );
         size += len;
+        boolean isSubArray = 0 != off || len < b.length ;
+        byte[] sub = ( isSubArray ? Arrays.copyOfRange( b, off, len ) : b );
         for ( final AbstractChecksumGenerator checksum : checksums )
         {
-            checksum.update( b );
+            checksum.update( sub );
```
ruhan1, line 147: I'd like to add a method update( final byte[] data, int offset, int len ) to AbstractChecksumGenerator, and call it here. This way we don't need to calc the sub-array by ourself. MessageDigest digester will offer it.

whitingjr: Ok let's do that.

PR #344, opened 2020-08-24
Add one attr in Location for directory listing handling

ligangty:


ligangty:
Merged as this is a very simple change.

PR #347, opened 2020-09-03
Galley transfers inject fields keyed to upstream hostname

ruhan1:
This is to wrap the download job's doCall with withStandardMetricWrapper. So that the inteceptor can catch the elapse time and send to Honeycomb.

### DIFF
--- a/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
+++ b/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
@@ -66,6 +66,13 @@ public final class HttpDownload
 
     public HttpDownload( final String url, final HttpLocation location, final Transfer target,
                          final Map<Transfer, Long> transferSizes, final EventMetadata eventMetadata, final Http http,
+                         final ObjectMapper mapper )
+    {
+        this( url, location, target, transferSizes, eventMetadata, http, mapper, true, null, null, null );
+    }
+
+    public HttpDownload( final String url, final HttpLocation location, final Transfer target,
+                         final Map<Transfer, Long> transferSizes, final EventMetadata eventMetadata, final Http http,
                          final ObjectMapper mapper, final MetricRegistry metricRegistry,
                          final TransportMetricConfig transportMetricConfig,
                          final HoneycombManager honeycombManager )
--- a/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
+++ b/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
@@ -16,7 +16,6 @@
 package org.commonjava.maven.galley.transport.htcli.internal;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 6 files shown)


transports/httpclient/src/test/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownloadTest.java
```diff
@@ -134,7 +134,7 @@ public void handle( final HttpServletRequest httpServletRequest,
 
         HttpDownload dl =
                 new HttpDownload( url, location, transfer, new HashMap<Transfer, Long>(), new EventMetadata(),
-                                  fixture.getHttp(), new ObjectMapper(), metricRegistry, metricConfig );
+                                  fixture.getHttp(), new ObjectMapper(), metricRegistry, metricConfig, null );
```
sswguo, line 137: How about adding another constructor in HttpDownload to avoid to pass null here? I see there are several ones here. :)

sswguo:
Looks good.

transports/httpclient/src/test/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownloadTest.java
ruhan1: This is good idea. In fact I wanted to remove the last 3 params which are not needed for test.

sswguo: Yeah, agree.

ruhan1: I refected some tests. pls review again.

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
```diff
@@ -128,6 +150,18 @@ public DownloadJob call()
         }
     }
 
+    // Generate a mangled name for metric/honeycomb. e.g., https://maven.apache.org to https.maven_apache_org
+    private String mangledName()
+    {
+        return name( this.request.getProtocolVersion().getProtocol(),
```
ligangty, line 156: Does this mean that all request from same host name with same protocol will be collected into same one metrics?

ligangty (commented):
Generally LGTM with one question below.

transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpDownload.java
ruhan1: Honeycomb data are collected per 'span', similar to per request. And we can query them and make a collective report, e.g., the download latency for a specific remote server.

ruhan1: So the short answer to the question is yes.

ligangty: Ok. Then no question here.

sswguo:
LGTM.

PR #348, opened 2020-09-07
Move request construct early to avoid npe

ruhan1:
The request object is used to create honeycomb name. Construct it early.


sswguo:
LGTM.

ligangty (commented):
LGTM

PR #351, opened 2020-10-21
Fix jdk11 compile error

ruhan1:
Basically the javax.annotation and javax.activation are missing on jdk11 so we add them explicitly.


ligangty (commented):
LGTM

sswguo:
LGTM!

ruhan1:
This is a dup. Will use Jeremy's pr.

PR #355, opened 2020-12-21
Fix IO stream resource leaks.

whitingjr:
This PR patches some situations where a resource is not closed. Or not closed in exceptional situations.

### DIFF
--- a/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/util/HttpUtil.java
+++ b/transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/util/HttpUtil.java
@@ -17,9 +17,14 @@ package org.commonjava.maven.galley.transport.htcli.util;
 
 import static org.apache.commons.io.IOUtils.closeQuietly;
 
+import java.io.IOException;
+import java.io.InputStream;
+
 import org.apache.http.Header;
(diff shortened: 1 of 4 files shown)


transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/util/HttpUtil.java
```diff
@@ -48,6 +53,11 @@ public static void cleanupResources( final CloseableHttpClient client, final Htt
             {
                 ( (AbstractExecutionAwareRequest) request ).reset();
             }
+            if ( request instanceof HttpPut )
+            {
+                HttpEntity e = ( (HttpPut)request ).getEntity();
+                EntityUtils.consumeQuietly( e );
+            }
```
ruhan1, line 60: It seems this cleanup is used in the finally clause of http requests. I am not quite sure why the cleanup is needed because the req/resp objects are ephemoral and will be GC-ed. But this looks harmless. If we do it for httpput, should we handle the httppost too?

whitingjr: I agree with you. After checking the callers they are all working with ephemeral streams. I'll close this PR.

PR #358, opened 2021-01-06
Bump junit to 4.13.1 for security fix

ligangty:
Seems junit 4.13 has some api changes on hamcrest level like below:
       * CoreMatchers.equalTo -> CoreMatchers.is
       * TemporaryFolder.newFolder(folder...) will throw IOException
       * Deprecated org.junit.Assert.assertThat to use org.hamcrest.MatcherAssert.assertThat


ruhan1 (commented):
LGTM

Issue #356, opened 2021-01-05
Adjust the identity value used for storing ConnectionManagerTracker in the cache.


whitingjr:
Currently the identity includes the artifact name. Which makes a connection cache not function as expected. See the [issue](https://github.com/Commonjava/indy/issues/1826) raised on the Indy project.

PR #359, opened 2021-01-06
Adjust identity of location to support connection pooling.

whitingjr:
This PR looks to solve #356 issue. Currently the identity of a location includes the artifact GAV.
 This is a problem as every http location has a unique identity. Which makes connection pooling not work as expected.
 A fix has been coded along with some unit tests to ensure a tracker per repository is stored in the cache.


whitingjr:
This change is considered unnecessary.

PR #363, opened 2021-03-18
Fix compilation and test error with o11yphant versin change

ligangty:


ligangty:
Merged directly as this is just a test fix.

PR #364, opened 2021-04-19
Adjust to refactored o11yphant trace API

jdcasey:
This will use the new jhttpc helper to inject client traces whenever files are downloaded, uploaded, listed, checked for existence, etc.

### DIFF
--- a/cdi-embedder/src/test/java/org/commonjava/maven/galley/embed/TestCDIProvider.java
+++ b/cdi-embedder/src/test/java/org/commonjava/maven/galley/embed/TestCDIProvider.java
@@ -15,7 +15,6 @@
  */
 package org.commonjava.maven.galley.embed;
 
-import io.opentelemetry.api.trace.Tracer;
 import org.commonjava.cdi.util.weft.config.DefaultWeftConfig;
 import org.commonjava.cdi.util.weft.config.WeftConfig;
 import org.commonjava.maven.galley.cache.partyline.PartyLineCacheProvider;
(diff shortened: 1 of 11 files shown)


geored (approved):
LGTM

cdi-embedder/src/test/java/org/commonjava/maven/galley/embed/TestCDIProvider.java
```diff
@@ -270,9 +306,9 @@ public String getNodeId()
 
     @Produces
     @Default
-    public DefaultTrafficClassifier getTrafficClassifier()
+    public AbstractTrafficClassifier getTrafficClassifier()
```
ligangty, line 309: For the naming here, not sure if an abstract class as result of a producer method is good option. I'm always feeling if an interface is more proper. But this is not a big deal.

ligangty (approved):
LGTM

cdi-embedder/src/test/java/org/commonjava/maven/galley/embed/TestCDIProvider.java
geored: Yes, i noticed that also and also default annotation is quastionable.

jdcasey: Hm, I would agree. I'll look at why it's not using the interface there.

jdcasey: I had to make a very minor change to o11yphant to resolve a remaining reference to the abstract classifier, in the GoldenSignalsFilter. I've done that here: https://github.com/Commonjava/o11yphant/pull/56

PR #371, opened 2021-10-27
Delete empty dir quietly during listing

ruhan1:
Sometimes after rollback a promotion, there is an empty dir left. But galley failed to delete it during the listing (in order to re-create the meta).
This is because nfs left a weird .nfsxxxx file there. After a while, it will be gone. We can just ignore such error during the listing.


ruhan1:
This is a minor fix. I merged it.

ruhan1:
btw, our own CI passed.

PR #372, opened 2021-11-10
Checking the timeout before returning transfer

sswguo:
Let's enable this and check if the file is timeout, to avoid to return the stale file. There are 1000+ stale npm package metadata files in devel-master,  and we have to delete one by one without this checker.


ruhan1 (approved):
LGTM

sswguo:
Merging this for test. thanks.

PR #373, opened 2021-12-22
Add log for investigation

sswguo:
I am investigating the issue: MMENG-2392. Where the logs showed that there were duplicated download jobs created when there were two requests for the same package. And I can reproduce it on stage but cannot on local. So adding this log to check the pending jobs to see if that helps.


ruhan1 (approved):
LGTM

PR #374, opened 2021-12-24
Fix pathmap cacheprovider timeout issue

ruhan1:
The old code use 86400s (24 hours) as the default timeout, and removes any files old than that. This is wrong.
1. "DEFAULT_TIMEOUT_SECONDS = -1". It mean never timeout. 
2. Refactor the provider constructor, remove the hard code config, take the config object as param so Indy can pass one to galley.
I add a test to verify the expiration behavior.


sswguo:
LGTM.

PR #377, opened 2021-12-28
Logs for investigation

sswguo:
I still cannot reproduce this locally while it always happens on stage instance, that even the target is already in the Map, it still creates new job.


sswguo:
I will merge this for test and this should be revert later. thanks.

PR #382, opened 2022-01-14
Improve the HttpDownload info log

ruhan1:
When download hit 404, indy still print:
Download attempt done: https://... Result:
  target: RepositoryLocation [npm:remote:npmjs]:... (stored at: /opt/indy/var/lib/indy/storage/...)
  error: null
This is confusing. I change the log to indicate failed or success downloads.


sswguo:
LGTM.

PR #387, opened 2022-05-30
Fix isTransferTimeout not working issue

ruhan1:
There are two bugs.
A. the Indy's config is not passed to PathMappedCacheProviderFactory properly, so the timeout check is never enabled. 
B. the path use to get special info not work for NPM. e.g, it passes the "/jquery" and the special checks are not aware of it. We need to translate it to "/jquery/package.json".


sswguo (approved):
LGTM.

PR #390, opened 2022-06-07
Check and remove empty dir in listing if ALLOW_REMOVE_EMPTY_DIRECTORY is true

ruhan1:
As a way of clean-up, this try to list all sub-folders to see if they are empty. e.g., if a dir has 1000 sub folders, it actually will do 1000 more listing operations. On some storage system (pathmap storage), this can cause serious performance problem.
The clean-up is useful to some extent. e.g, if there is a maven version folder and for some reason it is empty, the metadata generation will skip it. To fix the performance and still keep the ability of empty check, I use a flag in event metadata to indicate whether the listing need to filter out empty folders.


sswguo:
LGTM.

ligangty (commented):
LGTM

PR #391, opened 2022-07-11
Add log for removing resource

sswguo:
When I debug the issue: MMENG-3066, there is no related `DELETE` operation from outside, I am suspect that we may delete this as timeout, but there is no logs for this. Let's record the delete caused by the timeout, to help to investigate.


ruhan1 (commented):
LGTM

PR #392, opened 2022-07-22
Add timeoutProcessing support for partyline cache factory

yma88:


caches/partyline/src/main/java/org/commonjava/maven/galley/cache/partyline/PartyLineCacheProviderConfig.java
```diff
@@ -31,9 +31,10 @@
 
     private Integer defaultTimeoutSeconds;
 
-    public PartyLineCacheProviderConfig( final File cacheBasedir )
+    public PartyLineCacheProviderConfig( final File cacheBasedir, final Boolean timeoutProcessing )
```
ruhan1, line 34: There is a method 'withTimeoutProcessingEnabled'. maybe just update Indy side to call it?

caches/partyline/src/main/java/org/commonjava/maven/galley/cache/partyline/PartyLineCacheProvider.java
```diff
@@ -68,15 +68,15 @@
 
     private List<Transfer> toDelete = Collections.synchronizedList( new ArrayList<>() );
 
-    public PartyLineCacheProvider( final File cacheBasedir, final PathGenerator pathGenerator,
+    public PartyLineCacheProvider( final File cacheBasedir, final Boolean timeoutProcessing, final PathGenerator pathGenerator,
```
ligangty, line 71: Recommend to give another constructor with this param, and leave this one with a default value of this timeoutProcessing to call the new one. This can make the constructor compatible with other lib without any changes.(if possible, or if this change is really needed)

ligangty (commented):
I don't see any changes of this in partyline side. So why do we have this one?

caches/partyline/src/main/java/org/commonjava/maven/galley/cache/partyline/PartyLineCacheProviderConfig.java
yma88: @ruhan1 Indy uses PartyLineCacheProviderFactory to build, I update to use this when creating partyline CacheProvider instead

yma88:
> I don't see any changes of this in partyline side. So why do we have this one?

@ligangty I change to use withTimeoutProcessingEnabled to transfer timeoutProcessing value instead, and this is for Indy, https://github.com/Commonjava/indy/pull/2112

ligangty (commented):
LGTM

PR #394, opened 2022-07-28
Fix: Upgrade libs for security fix

ligangty:
* web-commons-bom: 24 -> 27 (for commons-io 2.4 -> 2.7)
   * commons-codec: 16 -> 1.15


ligangty:
lib upgrade, so merge directly

PR #399, opened 2022-11-28
Upgrade deps

ligangty:
* cdi and weld upgrade to cdi 2.0 spec
  * weft to 1.21-SNAPSHOT
  * web-commons-bom to 28-SNAPSHOT


ligangty:
Just version upgrade and ci passed, so merged directly.

PR #401, opened 2022-12-06
Code optimization: Fix bunch of IDE static check warnings

ligangty:


ligangty:
This is code optimization for IDE warnings. No logic changes so merged directly.

PR #405, opened 2022-12-29
Fix the timeout of the metadata resource

sswguo:
Try to fix the issue MMENG-3428: "NPM group metadata were not refreshed".  Ref: https://github.com/Commonjava/galley/blob/master/api/src/main/java/org/commonjava/maven/galley/util/LocationUtils.java#L54

the metadata under group should be refreshed regularly but not never timeout.


ligangty (commented):
LGTM

ruhan1 (commented):
LGTM

PR #406, opened 2023-02-16
Fix the checker on expiration

sswguo:
This includes:

- checksum files should be set `false` on metadata field, keep the NPM setting same with maven setting
- add the checker for resource before deletion, avoid to delete the files from the readonly repo

### DIFF
--- a/caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/PathMappedCacheProvider.java
+++ b/caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/PathMappedCacheProvider.java
@@ -240,7 +240,7 @@ public class PathMappedCacheProvider
     public Transfer getTransfer( final ConcreteResource resource )
     {
         Transfer txfr = new Transfer( resource, this, fileEventManager, transferDecorator );
-        if ( !resource.isRoot() && config.isTimeoutProcessingEnabled() )
+        if ( resource.allowsDeletion() && !resource.isRoot() && config.isTimeoutProcessingEnabled() )
         {
             if ( isTransferTimeout( txfr ) )
             {
--- a/api/src/main/java/org/commonjava/maven/galley/io/SpecialPathConstants.java
+++ b/api/src/main/java/org/commonjava/maven/galley/io/SpecialPathConstants.java
@@ -178,8 +178,6 @@ class NPMSpecialPathSet implements SpecialPathSet
 
         npmSpecialPaths.add( SpecialPathInfo.from( new FilePatternMatcher( ".*(\\.md5|\\.sha[\\d]+)$" ) )
                                             .setDecoratable( false )
-                                            .setMergable( true )
-                                            .setMetadata( true )
                                             .build() );
     }
 


caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/PathMappedCacheProvider.java
```diff
@@ -240,7 +240,7 @@ public String getStoragePath( final ConcreteResource resource )
     public Transfer getTransfer( final ConcreteResource resource )
     {
         Transfer txfr = new Transfer( resource, this, fileEventManager, transferDecorator );
-        if ( !resource.isRoot() && config.isTimeoutProcessingEnabled() )
+        if ( resource.allowsDeletion() && !resource.isRoot() && config.isTimeoutProcessingEnabled() )
         {
             if ( isTransferTimeout( txfr ) )
```
ruhan1, line 245: do you know why the checksum files are expired? If they are not treated as metadata, they should not expire, same as the normal files.

sswguo: checking this loop: https://github.com/Commonjava/galley/blob/master/caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/PathMappedCacheProvider.java#L279-L293
when checking with maven info, it's not metadata but it will check against NPM setting which will treat it as metadata. 
So we may need to find a way to refactor this loop as well.

ruhan1: +1, we should first know the pkgType and then use it to check.

ruhan1:
LGTM

yma88:
LGTM, another one place: Indy path producer also treats it as metadata: https://github.com/Commonjava/indy/blob/master/addons/pkg-npm/common/src/main/java/org/commonjava/indy/pkg/npm/data/NPMSpecialPathProducer.java

caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/PathMappedCacheProvider.java
ligangty: Just checked the KeydLocation and seems all of them are returning true except CacheOnlyLocation with readOnly=true. Does this mean the original bug deleted a readonly hosted repo's signature files?

sswguo: > Just checked the KeydLocation and seems all of them are returning true except CacheOnlyLocation with readOnly=true. Does this mean the original bug deleted a readonly hosted repo's signature files?

1. The original thing is metadata in group is never expired like normal artifacts. But in 3.0.1, I made the change to expire the metadata based on the repo's `metadata-timeout` setting ( if there is no such setting, using the default value with 4 days ),  but since NPM side treat the checksum files as metadata too, that's why it deleted that files too.
2. I found it deleted the checksum files in hosted repo too, that's caused by the missing checker here. And yeah, only CacheOnlyLocation has that field, and Indy side converts the Hosted repo into CacheOnlyLocation, ref: https://github.com/Commonjava/indy/blob/master/api/src/main/java/org/commonjava/indy/util/LocationUtils.java#L82-L85   which can pass the readonly attribute into galley model.

sswguo:
> LGTM, another one place: Indy path producer also treats it as metadata: https://github.com/Commonjava/indy/blob/master/addons/pkg-npm/common/src/main/java/org/commonjava/indy/pkg/npm/data/NPMSpecialPathProducer.java

Yes, I will make the change in indy side too. thanks.

caches/path-mapped/src/main/java/org/commonjava/maven/galley/cache/pathmapped/PathMappedCacheProvider.java
ligangty: Ok sounds good.

ligangty (approved):
LGTM

PR #417, opened 2023-10-18
Support 'path+query' base64 encode

ruhan1:
This is part 2 for [MMENG-3880](https://issues.redhat.com/browse/MMENG-3880) - Take query string into account when downloading through generic proxy.
If a Location set the attribute 'path-encode' : 'base64', the buildUrl() will decode the path string by base64 before sending to remote site.
Reason behind: sometimes the remote url comes with query params (?xx-xx) and this style does not match the natural way of caching the file to local storage. In this case, we make such Locations as base64 encoded. The caller encodes the path before sending http request to such Locations.


ligangty (commented):
LGTM

sswguo:
LGTM.

PR #430, opened 2024-08-08
Use GlobalProxyConfig as proxy bean to wrap location

yma96:


yma96:
Just change the proxy config bean to GlobalProxyConfig to avoid Indy CDI issue.

yma96:
Just merge this to fix the CDI problem of https://github.com/Commonjava/indy/pull/2450.

PR #435, opened 2024-08-20
Add proxy as alternative way to do transfer http jobs

yma96:


yma96:
Hi, could you have time to have a review on this? thanks.

ruhan1 (commented):
LGTM

ligangty:
Generally LGTM. Is it possible to add unit test for these new code?

yma96:
@ligangty Sure, some tests could cover from Indy side, I'll also try to cover more from Galley.
Merge this first to test in iad2, then make up UT and commit later.

PR #436, opened 2024-08-21
Use more elegant way to achieve the proxy retry and extend the default job wait timeout

yma96:


yma96:
Hi, could you help to review on this? I do some change on the code for proxy to make it more graceful and the default timeout for job needs to extend, you could checkout the comment in code line, thanks.

ruhan1 (commented):
LGTM

PR #444, opened 2025-01-20
Add Proxy Site cache support for squid proxy delivery

yma96:


transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/AbstractHttpJob.java
```diff
@@ -104,13 +105,25 @@ protected boolean executeHttp()
...
                 tries--;
                 try
                 {
+                    // if the site is in memory, go with proxy directly, avoid to query db frequently through isProxySite
+                    if ( getProxySitesCache().contains( site ) )
+                    {
+                        doProxy = true;
+                    }
+                    // if the site is not in memory, query db through isProxySite
+                    else if ( isProxySite( site ) )
```
sswguo, line 122: For those sites already in the whitelist, this will check it every request, right ? it's better to avoid that if it's the case.

yma96: @sswguo yeah, that's the point. I'll try to add another mem cache for whitelist, and any first successful request going to the site without proxy will add into it, the following access on this whitelist site will check the mem cache instead to avoid frequent db query.
I'm doing refactor on https://github.com/Commonjava/path-mapped-storage/pull/121, will amend this point together.

yma96:
Close this and will commit new PR.

PR #446, opened 2025-02-07
Add Proxy Site cache support for squid proxy delivery

yma96:


transports/httpclient/src/main/java/org/commonjava/maven/galley/transport/htcli/internal/HttpPublish.java
```diff
@@ -42,10 +43,10 @@ public final class HttpPublish
 
     private boolean success;
 
-    public HttpPublish( final String url, final HttpLocation location, final InputStream stream, final long length, final String contentType,
-                        final Http http )
+    public HttpPublish( final String url, final HttpLocation location, final InputStream stream, final long length,
+                        final String contentType, final Http http, ProxySitesCache proxySitesCache )
     {
-        super( url, location, http, HttpStatus.SC_OK, HttpStatus.SC_CREATED );
+        super( url, location, http, proxySitesCache, HttpStatus.SC_OK, HttpStatus.SC_CREATED );
```
sswguo, line 49: The publish one may not need this cache, we can pass NULL to match the method signature.

yma96: @sswguo Actually we have the final place GlobalProxyConfig to configure the allowable job types, while to avoid the unnecessary param assigned and cache validation, I remove this part, thanks, commit: https://github.com/Commonjava/galley/pull/446/commits/b2024993076e7f298697f6562220ac6828cb0ec8

sswguo:
LGTM.

